PR #510, opened 2014-01-16
More R-op and grad implemented for Alex's Cuda

pascanur:
Hi, 

 So I've implemented the grad of grad/ Rop of grad/grad of Rop for {maxpool, filter_acts}. I've improved (to some extend) the test as well. This code is needed by Yann's (and mine) smoothness regularization (or manifold counterpart). For now we can use our branch of pylearn2, but if you think the change is reasonable, you could also merge it pylearn2. It does not affect old code.


dwf:
This will never be merged as is, created #1400 to pick it up.

PR #528, opened 2014-01-27
New parameter initialization refactor (softmax, relu, maxout fully-connected)

dwf:


dwf:
After about my 10th attempt to rebase this against current master, I give up. MLP needs a massive refactor before this is worthwhile.

PR #666, opened 2014-03-05
Dataset tests

capybaralet:


dwf:
This door is opened elsewhere.

PR #860, opened 2014-04-29
Implements setter and getter for inv_P and P

bouthilx:
inv_P and P are used in zca_dataset and are accessed
directly. It also makes the assumption that a ZCA object has  
those attributes after initialization, but it's not true for old 
pickled ZCA. 

This pull request adds a getter/setter interface
to access them and also get rids of the assumption that
inv_P_ or P_ is an attribute of t.

It completes a fix made in #833

### DIFF
--- a/pylearn2/datasets/preprocessing.py
+++ b/pylearn2/datasets/preprocessing.py
@@ -40,6 +40,7 @@ from pylearn2.utils import contains_nan
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


dwf:
@bouthilx This LGTM, can you rebase?

@goodfeli If you have any objection to merging this let me know.

goodfeli:
Feel free to merge after rebase

PR #892, opened 2014-05-12
SVHN cache

### LINKED ISSUE #601: update SVHN dataset to new local dataset interface
When gh-96 is merged, we need to update the SVHN datasets and associated doc to use the new long term interface to local datasets.

memimo:
Fixes #601

Note to lisa members: Train data (.h5) on lisa PYLEARN2_DATA_PATH is broken and has to be processed again.

### DIFF
--- a/pylearn2/datasets/dense_design_matrix.py
+++ b/pylearn2/datasets/dense_design_matrix.py
@@ -1194,7 +1194,8 @@ class DenseDesignMatrixPyTables(DenseDesignMatrix):
                                             data_x=X,
                                             start=start)
 
-    def init_hdf5(self, path, shapes):
+    @staticmethod
+    def init_hdf5(path, shapes, filters=None):
         """
         Initializes the hdf5 file into which the data will be stored. This must
         be called before calling fill_hdf5.
@@ -1210,14 +1211,16 @@ class DenseDesignMatrixPyTables(DenseDesignMatrix):
         x_shape, y_shape = shapes
         # make pytables
         ensure_tables()
+        if filters is None:
+            filters = tables.Filters(complib='blosc', complevel=5)
         h5file = tables.openFile(path, mode="w", title="SVHN Dataset")
         gcolumns = h5file.createGroup(h5file.root, "Data", "Data")
         atom = (tables.Float32Atom() if config.floatX == 'float32'
                 else tables.Float64Atom())
         h5file.createCArray(gcolumns, 'X', atom=atom, shape=x_shape,
-                            title="Data values", filters=self.filters)
+                            title="Data values", filters=filters)
         h5file.createCArray(gcolumns, 'y', atom=atom, shape=y_shape,
-                            title="Data targets", filters=self.filters)
+                            title="Data targets", filters=filters)
         return h5file, gcolumns
 
     @staticmethod
--- a/pylearn2/datasets/svhn.py
+++ b/pylearn2/datasets/svhn.py
@@ -15,6 +15,7 @@ import numpy
 from theano.compat.six.moves import xrange
 from theano import config
 from pylearn2.datasets import dense_design_matrix
+from pylearn2.datasets import cache
 from pylearn2.utils.serial import load
 from pylearn2.utils.string_utils import preprocess
 from pylearn2.utils.rng import make_np_rng
@@ -29,14 +30,27 @@ class SVHN(dense_design_matrix.DenseDesignMatrixPyTables):
 
     Parameters
     ----------
-    which_set : WRITEME
-    path : WRITEME
-    center : WRITEME
-    scale : WRITEME
-    start : WRITEME
-    stop : WRITEME
-    axes : WRITEME
-    preprocessor : WRITEME
+    which_set : str
+        Subset name.
+    path : str, optional
+        Path to pytable h5 files in case you want to use
+        preprocessed/modified data.
+    center : bool, optional
+        Center data around zero.
+    scale : bool, optional
+        Scale data to [0, 1] from [0, 255]
+    start : bool, optional
+        Start index
+    stop : bool, optional
+        Stop index
+    axes : str, tuple
+        Axes format.
+    preprocessor : obj, optional
+        Preprocessor
+    enable_cache : bool, optional
+        enables pylearn2 local caching.
+    write_permission: bool, optional
+        give write permission to the files. Use this with caution.
     """
 
     mapper = {'train': 0, 'test': 1, 'extra': 2, 'train_all': 3,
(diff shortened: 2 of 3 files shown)


pylearn2/datasets/dense_design_matrix.py
```diff
@@ -1227,7 +1227,8 @@ def set_topological_view(self, V, axes=('b', 0, 1, 'c'), start=0):
                                             data_x=X,
                                             start=start)
 
-    def init_hdf5(self, path, shapes):
+    @staticmethod
+    def init_hdf5(path, shapes, filters=None):
```
memimo: The reason is that, I want to make the first time data construction separated from its usage.

memimo:
This is very specific kind of pre-processing for one paper. I don't think there is any need to waste disk space in the PYLEARN2_DATA_PATH
As it was discussed in our last meeting, we agreed to less focus on fully supporting papers specific code and put most of our efforts in the library's core functionalities.

pylearn2/datasets/svhn.py
```diff
@@ -467,7 +487,7 @@ def split_train_valid(path, num_valid_train=400,
         # The original splits
         if which_set in ['train', 'test']:
             data_x, data_y = load_data("{0}{1}_32x32.mat".format(path,
-                                                                 which_set))
+                                       which_set))
```
memimo: Are you sure? autopep8 seems to be disagreeing.

memimo:
I applied the other changes you asked.

pylearn2/datasets/svhn.py
```diff
@@ -29,14 +30,28 @@ class SVHN(dense_design_matrix.DenseDesignMatrixPyTables):
...
-    which_set : WRITEME
-    path : WRITEME
-    center : WRITEME
-    scale : WRITEME
-    start : WRITEME
-    stop : WRITEME
-    axes : WRITEME
-    preprocessor : WRITEME
+    which_set : str
+                Subset name.
```
dwf: Indents should be 4 spaces according to the docstring standard. Also a list of valid values here (I assume 'train', 'test', 'valid', but aren't there some weird ones?).

pylearn2/datasets/svhn.py
```diff
@@ -29,14 +30,28 @@ class SVHN(dense_design_matrix.DenseDesignMatrixPyTables):
...
+    path : str, optional
+           Path to pytable h5 files in case you want to use
+           preprocessed/modified data.
+    center : bool, optional
+             Center data around zero.
```
dwf: How is this done? On a per-image or per batch basis? I'm assuming you're not actually subtracting the mean of the hundreds of thousands of images...

pylearn2/datasets/svhn.py
```diff
@@ -29,14 +30,28 @@ class SVHN(dense_design_matrix.DenseDesignMatrixPyTables):
...
+    scale : bool, optional
+            Scale data to [0, 1]
```
dwf: Mention what the original scale is?

pylearn2/datasets/svhn.py
```diff
@@ -29,14 +30,28 @@ class SVHN(dense_design_matrix.DenseDesignMatrixPyTables):
...
+    start : bool, optional
+            Start index
+    stop : bool, optional
+           Stop index
+    axes : str, optional
+           Axes format.
```
dwf: This is str?

pylearn2/datasets/svhn.py
```diff
@@ -29,14 +30,28 @@ class SVHN(dense_design_matrix.DenseDesignMatrixPyTables):
...
+    preprocessor : obj, optional
+                   Preprocessor
+    enable_cache : bool, optional
+                   enables pylearn2 local caching.
+    write_permission :
```
dwf: bool, optional

pylearn2/datasets/svhn.py
```diff
@@ -46,44 +61,48 @@ class SVHN(dense_design_matrix.DenseDesignMatrixPyTables):
 
     def __init__(self, which_set, path=None, center=False, scale=False,
                  start=None, stop=None, axes=('b', 0, 1, 'c'),
-                 preprocessor = None):
+                 preprocessor = None, enable_cache=True,
```
dwf: PEP8, no space around =

pylearn2/datasets/svhn.py
```diff
@@ -46,44 +61,48 @@ class SVHN(dense_design_matrix.DenseDesignMatrixPyTables):
...
+                 write_permission=False):
 
         assert which_set in self.mapper.keys()
 
         self.__dict__.update(locals())
         del self.self
 
-        if path is None:
-            path = '${PYLEARN2_DATA_PATH}/SVHN/format2/'
+        if write_permission and (not enable_cache):
```
dwf: no need for the () but ok

pylearn2/datasets/svhn.py
```diff
@@ -46,44 +61,48 @@ class SVHN(dense_design_matrix.DenseDesignMatrixPyTables):
...
+            if path is None:
+                raise ValueError("You are not allowed to change the original" +
+                                 " file located at $PYLEARN2_DATA_PATH. " +
+                                 " Enable caching to remove this error")
```
dwf: period at the end.

pylearn2/datasets/svhn.py
```diff
@@ -46,44 +61,48 @@ class SVHN(dense_design_matrix.DenseDesignMatrixPyTables):
...
+            else:
+                warnings.warn("You might change the data in the files " +
+                              "located at: {}".format(path))
```
dwf: Under what circumstances? Can you be more specific? Will this class ever try and alter them?

pylearn2/datasets/svhn.py
```diff
@@ -46,44 +61,48 @@ class SVHN(dense_design_matrix.DenseDesignMatrixPyTables):
...
-
-        if mode == 'r' and (scale or
-                            center or
-                            (start is not None) or
-                            (stop is not None)):
-            raise ValueError("Only for speed there is a copy of hdf5 file in "
-                             "PYLEARN2_DATA_PATH but it meant to be only "
-                             "readable. If you wish to modify the data, you "
-                             "should pass a local copy to the path argument.")
+            warnings.warn("Mode is set to write more, any changes you apply " +
```
dwf: By which you mean... if the user modifies `X` and `y` attributes?

pylearn2/datasets/svhn.py
```diff
@@ -46,44 +61,48 @@ class SVHN(dense_design_matrix.DenseDesignMatrixPyTables):
...
-        # load data
         path = preprocess(path)
         file_n = "{0}_32x32.h5".format(os.path.join(path, "h5", which_set))
-        if os.path.isfile(file_n):
-            make_new = False
-        else:
-            make_new = True
-            warnings.warn("Over riding existing file: {0}".format(file_n))
+        if not os.path.isfile(file_n):
+            raise IOError("Pytable h5 file do not exist at {}.".format(path) +
```
dwf: PyTables* with an s

pylearn2/datasets/svhn.py
dwf: It looks like centering and scaling does modify data on disk. Same with preprocessor? Warnings should specify that.

pylearn2/datasets/svhn.py
```diff
@@ -214,11 +233,11 @@ def split_train_valid(path, num_valid_train=400,
             train_index = list(train_index)
 
             train_x = numpy.concatenate((train_x,
-                                         data['X'][:, :, :, train_index]),
+                                        data['X'][:, :, :, train_index]),
                                         axis=3)
             train_y = numpy.concatenate((train_y, data['y'][train_index, :]))
             valid_x = numpy.concatenate((valid_x,
-                                         data['X'][:, :, :, valid_index]),
+                                        data['X'][:, :, :, valid_index]),
```
dwf: Are you storing as one-hot instead of int? How does that interact with `IndexSpace` stuff (CC/@bartvm)

pylearn2/datasets/svhn.py
memimo: There is no official valid. There original sets are: train, extra, and test.
train_all: train + extra
splitted_train: train + extra - valid

pylearn2/datasets/svhn.py
memimo: data.X[:] -= 127.5

pylearn2/datasets/svhn.py
memimo: When path is passed, yes.

pylearn2/datasets/svhn.py
memimo: No, int. It does not support IndexSpace.

pylearn2/datasets/svhn.py
memimo: It won't got there, it will stop before here if scale or centering is set.

pylearn2/datasets/svhn.py
lamblin: I don't understand. If `y` consists in one-hot vectors, it would be in VectorSpace, but if it contains the indices, then it should be in IndexSpace.
Or are you saying that this conversion happens outside of the iterator?

memimo: Here is the story about one_hot. It did not support it originally, but someone added the functionality without testing and that's why it's broken. There is a separate PR #1371 to fix the issue

PR #916, opened 2014-05-19
Add support for cuda-convnet's filter groups

f0k:
This PR adds support for the `numGroups` argument to the convolution routines of Alex Krizhevsky's cuda-convnet. This is a particular kind of restricted connectivity that splits up the input channels and filters into same-sized groups, than applies each group of filters to its associated group of input channels only. This is termed ["block sparse convolution" in the cuda-convnet documentation](http://code.google.com/p/cuda-convnet/wiki/LayerParams#Block_sparse_convolution_layer).

The PR also prepares the interface for supporting ["random sparse convolution"](http://code.google.com/p/cuda-convnet/wiki/LayerParams#Random_sparse_convolution_layer) via the `dColorIndices` argument to Alex's convolution routines, but does not attempt to implement it.

Finally, the PR includes a test that compares the grouped convolution results of cuda-convnet with results using theano's builtin `conv2d` op (both for the outputs and gradients). Results are close enough to infer that the two implementations attempt to compute the same thing, just with different rounding errors.

One minor thing I'm unsure about: Is the `self.dense_connectivity` field of `BaseActs` used anywhere? I've left it in for now, but it doesn't play any role in the implementation. Let me know if you'd like me to remove it or have any other concerns.

Cheers, Jan

### DIFF
--- a/pylearn2/devtools/tests/test_format.py
+++ b/pylearn2/devtools/tests/test_format.py
@@ -438,6 +438,7 @@ whitelist_docstrings = [
     'sandbox/cuda_convnet/tests/test_pool.py',
     'sandbox/cuda_convnet/tests/test_common.py',
     'sandbox/cuda_convnet/tests/test_stochastic_pool.py',
+    "sandbox/cuda_convnet/tests/test_grouped_conv.py",
     'sandbox/cuda_convnet/shared_code.py',
     'sandbox/cuda_convnet/__init__.py',
     'sandbox/cuda_convnet/img_acts.py',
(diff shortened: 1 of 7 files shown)


dwf:
@nouiz Can you take another look at this?

f0k:
Hey, you're alive :) I know that everything was working at the time I did it, but I wouldn't be too excited about rebasing it 8 months later. I'd be glad if somebody could revive and merge this, though, and happy to answer any questions!

dwf:
Okay, I'm closing and filing a new ticket to come resurrect this.

PR #957, opened 2014-06-04
Optional MonitorBased logging.

dwf:
Displays a log message every time the countdown resets and also every `logging_period` epochs that an improvement hasn't been seen. Off by default.

### DIFF
--- a/pylearn2/termination_criteria/__init__.py
+++ b/pylearn2/termination_criteria/__init__.py
@@ -10,8 +10,12 @@ __maintainer__ = "LISA Lab"
 __email__ = "pylearn-dev@googlegroups"
 
 import functools
+import logging
 import numpy as np
 import warnings
+
+
+log = logging.getLogger(__name__)
 
 
 class TerminationCriterion(object):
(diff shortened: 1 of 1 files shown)


pylearn2/termination_criteria/__init__.py
```diff
@@ -58,18 +62,27 @@ class MonitorBased(TerminationCriterion):
         The threshold factor by which we expect the channel value to have
         decreased
     N : int
```
lamblin: Should be "int, optional", while you're at it.

pylearn2/termination_criteria/__init__.py
```diff
@@ -58,18 +62,27 @@ class MonitorBased(TerminationCriterion):
...
-        Number of epochs to look back
+        Number of epochs to look back. 5 by default.
     channel_name : string, optional
         Name of the channel to examine. If None and the monitor
         has only one channel, this channel will be used; otherwise, an
         error will be raised.
+    logging_period : int, optional
+        Log a message when the countdown is a multiple of this value.
+        If 0, disable all logging.
```
lamblin: Mention 0 is the default.

pylearn2/termination_criteria/__init__.py
```diff
@@ -112,6 +128,14 @@ def continue_learning(self, model):
         # The optimization continues until the countdown has reached 0,
         # meaning that N epochs have passed without the model improving
         # enough.
+        time_to_log = (self._logging_period > 0 and
+                       self.countdown % self._logging_period == 0)
+        if time_to_log:
+            # filler = "exceeds"
+            filler = "drops below"  # TODO: change when have higher_is_better
```
lamblin: Not sure what the TODO means. When would that be set?

lamblin:
Other than that, I'm OK with the change.

PR #1007, opened 2014-06-30
Feature/add stanford sentiment treebank dataset

trunghlt:
This branch added a generic interface to deal with text data and an initial dataset (Sentiment Treebank) which is a popular dataset for sentiment analysis.


dwf:
@bartvm You already had a look at this so can you give an update?

bartvm:
Would need a pretty thorough rewrite to play nice with the `TextDatasetMixin` class (https://github.com/lisa-lab/pylearn2/blob/master/pylearn2/sandbox/nlp/datasets/text.py) which was added around the same time. 

It also seems to have never passed Travis, currently because the function `_get_DATA_ROOT` returns `None` and makes `os.path.join` crash instead of failing gracefully when `PYLEARN2_DATA_PATH` is not set (which it isn't on Travis; so the tests should also be skipped using `skip_if_no_data`).

tl;dr Would need work to be merged currently, and more than I have time to patch up right now.

dwf:
Alright, closing for now, @trunghlt feel free to open again if you want to address @bartvm's comments.

Issue #677, opened 2014-03-08
Implement nesterov momentum


goodfeli:
TODO: implement Nesterov momentum. Easiest way to do it is via equivalence
between regular momentum and Nesterov momentum described in this note from
Nicolas Boulanger-Lewandowski:

Yes, I found that the following change of variable simplifies the implementation of Nesterov momentum.
It is in the same form as regular momentum in the sense that both velocity and parameter updates depend
only on the gradient at the current value of the parameters.

In short:

regular momentum:
(1) v_t = mu \* v_t-1 - lr \* gradient_f(params_t)
(2) params_t = params_t-1 + v_t
(3) params_t = params_t-1 + mu \* v_t-1 - lr \* gradient_f(params_t-1)

Nesterov momentum:
(4) v_t = mu \* v_t-1 - lr \* gradient_f(params_t-1 + mu \* v_t-1)
(5) params_t = params_t-1 + v_t

alternate formulation for Nesterov momentum:
(6) v_t = mu \* v_t-1 - lr \* gradient_f(params_t-1)
(7) params_t = params_t-1 + mu \* v_t - lr \* gradient_f(params_t-1)
(8) params_t = params_t-1 + mu**2 \* v_t-1 - (1+mu) \* lr \* gradient_f(params_t-1)

So with Theano you can use (1) then either (2) or (7)/(8) to have both options.

PR #1030, opened 2014-07-16
Nesterov momentum implementation

madisonmay:
Based off of the [simplification suggested by Ian Goodfellow](https://github.com/lisa-lab/pylearn2/issues/677) (@goodfeli) and [the corresponding paper by Yoshua Bengio](http://arxiv.org/pdf/1212.0901v2.pdf).

Still need to add tests to the test suite, but I'm looking for a few more eyes on the code to make sure the basics check out.

### DIFF
--- a/pylearn2/devtools/tests/test_format.py
+++ b/pylearn2/devtools/tests/test_format.py
@@ -13,7 +13,6 @@ from pylearn2.devtools.tests.pep8.pep8 import StyleGuide
 
 whitelist_pep8 = [
     "rbm_tools.py",
-    "training_algorithms/tests/test_learning_rule.py",
     "distributions/mnd.py",
     "models/sparse_autoencoder.py",
     "models/tests/test_dbm.py",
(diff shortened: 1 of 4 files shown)


obilaniu:
@lamblin I believe I've found a mistake in this implementation of Nesterov momentum. Neither the "alternate" rules given in PR #677 nor the rule given by madison are incorrect by my tracing of [Ilya Sutskever's paper](http://www.cs.utoronto.ca/~ilya/pubs/2013/1051_2.pdf), which gives the original formulation as equations (3) and (4) on page 3. Here are my proposed rules, with detailed commentary to assure everyone of their correctness.

nesterov
[Nesterov.pdf](https://github.com/lisa-lab/pylearn2/files/190821/Nesterov.pdf)

If confirmed it would mean that Pylearn never actually implemented Nesterov momentum correctly, and nobody discovered this in over one and a half years. An error this serious merits a notice to users to redo their experiments with NAG if they've previously been disappointed by it.

obilaniu:
I now think I was myself mistaken in my post above about a bug in the Nesterov momentum formulas, so I'm scrapping it, and you can ignore it.

Issue #881, opened 2014-05-08
checkpoint/resume of training.


nouiz:
Ian told that at least this need to be done:
- fix pickling of monitor
- make the train object resumable
- check the sgd class.

PR #1049, opened 2014-07-28
Pickling framework for the class 'Train'

AlassaneNdiaye:
#881


dwf:
Doesn't seem usable as it pickles the whole dataset.

PR #1061, opened 2014-08-04
Corrected small errors detected by pyflakes.

AlassaneNdiaye:


lamblin:
Newer versions of pyflakes do not consider that as errors, that is why it is not failing on Travis.
We just need to update the version of pyflakes used for the daily buildbot.

nouiz:
For the record, I did that update.

PR #1076, opened 2014-08-15
Update large_data.txt

skearnes:
Some updates to make the distinctions between PyTables and h5py clearer.


dwf:
Seems like this fell through the cracks, but good to merge AFAICT.

PR #1110, opened 2014-08-27
generalization of CSV dataset + shuffle feature

denadai2:
Rebased + added some features: https://github.com/lisa-lab/pylearn2/pull/1072

### DIFF
--- a/pylearn2/datasets/csv_dataset.py
+++ b/pylearn2/datasets/csv_dataset.py
@@ -3,25 +3,23 @@
 A simple general csv dataset wrapper for pylearn2.
 Can do automatic one-hot encoding based on labels present in a file.
 """
-__authors__ = "Zygmunt Zając"
+__authors__ = "Zygmunt Zając, Marco De Nadai"
 __copyright__ = "Copyright 2013, Zygmunt Zając"
-__credits__ = ["Zygmunt Zając", "Nicholas Dronen"]
+__credits__ = ["Zygmunt Zając", "Nicholas Dronen", "Marco De Nadai"]
 __license__ = "3-clause BSD"
 __maintainer__ = "?"
 __email__ = "zygmunt@fastml.com"
 
-import csv
 import numpy as np
-import os
 
 from pylearn2.datasets.dense_design_matrix import DenseDesignMatrix
-from pylearn2.utils import serial
 from pylearn2.utils.string_utils import preprocess
 
 
 class CSVDataset(DenseDesignMatrix):
 
-    """A generic class for accessing CSV files
+    """
+    A generic class for accessing CSV files
     labels, if present, should be in the first column
     if there's no labels, set expect_labels to False
     if there's no header line in your file, set expect_headers to False
@@ -57,6 +55,10 @@ class CSVDataset(DenseDesignMatrix):
 
     end_fraction : float
       The fraction of rows, starting at the end of the file, to load.
+
+    num_outputs : int, optional
+        number of target variables. defaults to 1
+
     """
     def __init__(self,
                  path='train.csv',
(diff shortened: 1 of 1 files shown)


lamblin:
@denadai2 If you are still interested in getting these changes merged, can you rebase on top of the current master? I'll have a new look after that.

denadai2:
ahahan not bad after 4 months
Now I'm rather busy, I'll rebase ASAP

evanhaldane:
I was about to submit a pull request allowing for multiple targets. What's the status of this?

denadai2:
@evanhaldane @lamblin done :)

pylearn2/datasets/csv_dataset.py
```diff
@@ -142,32 +151,35 @@ def _load_data(self):
...
+        if self.shuffle:
+            np.random.shuffle(data)
 
         if self.expect_labels:
-            y = data[:, 0]
-            X = data[:, 1:]
-            y = y.reshape((y.shape[0], 1))
+            y = data[:, 0:self.num_outputs]
+            x = data[:, self.num_outputs:]
+            y = y.reshape((y.shape[0], self.num_outputs))
```
lamblin: This part makes sense, thanks.

pylearn2/datasets/csv_dataset.py
```diff
@@ -142,32 +151,35 @@ def _load_data(self):
...
-                X = X[self.start:self.stop, ]
+                x = x[self.start:self.stop, ]
                 if y is not None:
                     y = y[self.start:self.stop]
 
-            return X, y
+            return x, y
+
+        if self.shuffle:
+            np.random.shuffle(data)
```
lamblin: This part is not needed, we can simply use "shuffled_sequential" as a mode to iterate over a shuffled version.
Also, it is problematic because it is not reproducible, we cannot control which permutation will be applied.
So I would get rid of the "shuffle" option.

pylearn2/datasets/csv_dataset.py
```diff
@@ -119,12 +128,12 @@ def __init__(self,
...
-        X, y = self._load_data()
+        x, y = self._load_data()
 
         if self.task == 'regression':
-            super(CSVDataset, self).__init__(X=X, y=y, **kwargs)
+            super(CSVDataset, self).__init__(X=x, y=y, **kwargs)
         else:
-            super(CSVDataset, self).__init__(X=X, y=y,
+            super(CSVDataset, self).__init__(X=x, y=y,
                                              y_labels=np.max(y) + 1, **kwargs)
```
lamblin: I don't understand why you had to change `X` to `x`, since the parent class uses `X`.

lamblin:
Thanks for the rebase!

pylearn2/datasets/csv_dataset.py
denadai2: because Pycharm was suggesting me that it is not PEP8, and I had PEP8 errors in travis :P shall I fix it with uppercase or lowercase x?

pylearn2/datasets/csv_dataset.py
denadai2: ok

pylearn2/datasets/csv_dataset.py
lamblin: OK, thanks for the explanation.
I think single-letter uppercase is not an issue at least for the pep8 script we are using.
It is not necessary that you revert everything back to capital X, but I think it would avoid some confusion.

evanhaldane:
thanks folks! :+1:

denadai2:
Ok I finished it. @lamblin @evanhaldane

lamblin:
Great, thanks!

denadai2:
thanks

PR #1111, opened 2014-08-27
Added the LMLS cost function and generalized the cost function

denadai2:
I generalized the cost function: I believe it's better for everyone, who can customize the behaviour of the mlp

I added LMLS: http://ieeexplore.ieee.org/xpl/login.jsp?tp=&arnumber=478411&url=http%3A%2F%2Fieeexplore.ieee.org%2Fxpls%2Fabs_all.jsp%3Farnumber%3D478411


dwf:
Seemingly abandoned, see above re: obscurity. Closing for now.

PR #1070, opened 2014-08-10
Added channel monitors to log/compute during the training

denadai2:
https://github.com/lisa-lab/pylearn2/issues/1068 it shows the list of possible monitoring channels and, if the user doesn't specify anything, pylearn2 works as it works now, otherwise it prints/calculate only the log_monitors passed.
It solves also the problem of the slowness of ipython.

Example: 

    !obj:pylearn2.train.Train {
    dataset: &train !obj:pylearn2.datasets.csv_dataset.CSVDataset {
        path: 'train.csv'
    },
    model: !obj:pylearn2.models.mlp.MLP {
        layers: [
                 !obj:pylearn2.models.mlp.Sigmoid {
                     layer_name: 'h0',
                     dim: 300,
                     sparse_init: 15
                 },
                 !obj:pylearn2.models.mlp.Linear {
                     layer_name: 'y',
                     irange: 0.5,
                     dim: 1
                 }
                ],
        nvis: 20,
    },
    algorithm: !obj:pylearn2.training_algorithms.sgd.SGD {
        batch_size: 10,
        learning_rate: .001,
        monitoring_dataset:
            {
                'train' : *train,
                'test'  : !obj:pylearn2.datasets.csv_dataset.CSVDataset {
                                path: 'testY.csv'
                            },
            },
        monitoring_channels: ['train_objective'],
        termination_criterion: !obj:pylearn2.termination_criteria.EpochCounter {
            max_epochs: 940
        },
        
    },
    extensions: [
        !obj:pylearn2.train_extensions.best_params.MonitorBasedSaveBest {
            channel_name: "test_objective",
            save_path: "best.pkl"
        }
    ],
    save_path: "softmax_regression.pkl",
    save_freq: 1
}

PR #1114, opened 2014-08-28
New monitoring system with enviromental vars

denadai2:
Ok I really need help here. Ref: https://github.com/lisa-lab/pylearn2/pull/1070/files

Anyway I added the configuration system of Theano. I'll propose them a pull request to generalize it. In configparser there is "SemicolonParam", which is a class that in the future should check the format of the parameter (which is a semicolon separated param).

What do you think so far?

Example of use:

    PYLEARN2_FLAGS='channels_include=train*' python ../../pylearn2/pylearn2/scripts/train.py NN_static2.yaml

### DIFF
--- a/pylearn2/__init__.py
+++ b/pylearn2/__init__.py
@@ -10,4 +10,4 @@ configure_custom()
 # Remove this from the top-level namespace.
 del configure_custom
 
-
+from configdefaults import pylearn2_config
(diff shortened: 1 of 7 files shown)


dwf:
This seems abandoned. Feel free to re-open if there's movement.

PR #1153, opened 2014-09-28
Add 'channels' option to enable non-interactive plotting

johnarevalo:
Include option to generate monitoring plots in a non-interactive environment like a bash script or ssh remote executions.


dwf:
This needs a rebase but LGTM otherwise.

lamblin:
Rebase has been done.
@johnarevalo The maintainers are not automatically notified when new commits are pushed, but they are if a new comment is submitted. Don't hesitate to add a comment when you update a PR.

johnarevalo:
@lamblin Thanks for your advice. I hope to do so in future PRs.

nouiz:
Note, we don't support Pylearn2 anymore and we are dropping Theano: 

https://groups.google.com/forum/#!topic/theano-users/7Poq8BZutbY

So I just want to war people that probably nobody will review this PR.

PR #1197, opened 2014-10-31
Add plotting extension

laurent-dinh:
Plotting extension : create png files containing plots for chosen monitoring channels.

### DIFF
--- a/pylearn2/train_extensions/plots.py
+++ b/pylearn2/train_extensions/plots.py
@@ -1,0 +1,356 @@
+"""
+Plot monitoring extensions while training.
+"""
+__authors__ = "Laurent Dinh"
+__copyright__ = "Copyright 2014, Universite de Montreal"
+__credits__ = ["Laurent Dinh"]
+__license__ = "3-clause BSD"
+__maintainer__ = "Laurent Dinh"
+__email__ = "dinhlaur@iro"
+
+
+import logging
+import os
+import os.path
+import stat
+import numpy
+np = numpy
+from pylearn2.train_extensions import TrainExtension
+
+from theano.compat.six.moves import xrange
+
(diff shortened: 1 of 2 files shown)


pylearn2/train_extensions/plots.py
```diff
@@ -0,0 +1,455 @@
...
+        Parameters
+        ----------
+        model : pylearn2.models.Model
+            The model trained
+        dataset : pylearn2.datasets.Dataset
+            The dataset on which the model is trained
+        algorithm : pylearn2.training_algorithms.TrainingAlgorithm
+            The algorithm the model is trained with
+
+        """
```
lamblin: Unfortunately, it is not possible.

lamblin:
Regarding the back-end selection, instead of always using Agg, you can check whether the `DISPLAY` environment variable is set: http://stackoverflow.com/questions/8257385/automatic-detection-of-display-availability-with-matplotlib
If it is not set, then use Agg, otherwise keep the default.

laurent-dinh:
Updated.

pylearn2/train_extensions/plots.py
```diff
@@ -0,0 +1,451 @@
...
+        raise NotImplementedError(str(type(self))+" does not implement setup.")
+
+    def plot(self):
+        """
+        Plot what it should.
```
lamblin: ಠ_ಠ

lamblin:
Thanks, just let me know when you have updated the documentation of `plot`, and manually checked that we have the desired behaviour:
- on a local machine with X, the default back-end is used
- on a remote connection with X (`ssh -X`), the default back-end is used
- on a headleass remote connection, the "Agg" back-end is used.

lamblin:
Thanks!

PR #1227, opened 2014-11-11
Simple change to allow for on-line data augmentation

cancan101:
See: https://groups.google.com/d/topic/pylearn-users/LLlBpl9z0MY/discussion

I am using this like this:
```python
class MyDenseDesignMatrix(DenseDesignMatrix):
    def get_topo_batch_transform_fn(self, fn):
        def impl(batch):
            batch = fn(batch)
            batch = np.array([affine_transform_rnd(b) for b in batch])
            return batch[..., np.newaxis]
        return impl
```

### DIFF
--- a/pylearn2/datasets/dense_design_matrix.py
+++ b/pylearn2/datasets/dense_design_matrix.py
@@ -266,6 +266,9 @@ class DenseDesignMatrix(Dataset):
             assert self.y.ndim <= 2
             assert np.all(self.y < self.y_labels)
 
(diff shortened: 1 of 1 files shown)


dwf:
Any problem with merging this?

lamblin:
This PR adds an additional hook which should not be necessary, as it could be part of the ViewConverter, or in a separate Block used in TransformerDataset.
Once gh-1257 is fixed, there should be no need for it.

PR #1260, opened 2014-11-20
Fixes on RemoveMean

gaa-cifasis:


dwf:
@lamblin says this needs to be solved another way.

PR #1288, opened 2014-11-28
Added unit tests for rotated mnist and adult

kelvinxu:
Added two tests based on previous pull request #666 

Kelvin

### DIFF
--- a/pylearn2/datasets/tests/__init__.py
+++ b/pylearn2/datasets/tests/__init__.py
@@ -1,0 +1,3 @@
+"""
+Dataset testing classes
+"""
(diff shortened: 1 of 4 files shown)


kelvinxu:
added docstrings/ checked files with pyflakes

memimo:
Travis is failing still because multiple docstring errors

kelvinxu:
I'll fix.

memimo:
Ideally we need more through tests epescially for mnist. But since it's been used less we can go as it is for now.
One more comment is, if you're fixing minor issues you can avoid making new commits by using git --amend.
Would you mind cleaning your commit history with git rebase. Git commits should be divided based on the logic not the author's work progress. You need to force push after you clean the history, then it should be ready for merge.

kelvinxu:
Thanks for the comments, I'll keep that in mind in the future. I squashed a bunch of the commits that were just docstrings/pep8.

memimo:
Thx!

PR #1290, opened 2014-12-01
Initial work on issue 921: creating tests for FlattenerLayer.

julianser:
I need some feedback on the following tests for FlattenerLayer:
1) I can't get the fprop to work on my toy example, what is wrong?
2) Do we need more tests on the "inner" FlattenerLayer to see if its other methods (e.g. get_monitoring_channels, get_weights, cost...) agree with what they are truly supposed to be? @dwf


lamblin:
Thanks!
Now that you have fixed the pep-8 formatting, can you please remove the file from the whitelist_pep8 at https://github.com/lisa-lab/pylearn2/blob/master/pylearn2/devtools/tests/test_format.py#L154 ?
Sorry I wasn't clear about that earlier.

PR #1320, opened 2014-12-20
added GRU

jych:
added Gated Recurrent Unit based on Cho-et-al 2014 EMNLP.


dwf:
@jych It's a good idea to separate bug fix commits if you can, if not as a separate pull request then at least a separate commit; that way it's easier to notice when a pull request contains a bug fix of some urgency. :)

PR #1334, opened 2015-01-06
Add a requirements file for readthedocs.org

abergeron:


abergeron:
This is non-code and needed right now, so I merge it right away.

abergeron:
Or not since I can't

PR #1337, opened 2015-01-09
Implement undo_format_as fixes #1301

### LINKED ISSUE #1301: Bug in FlattenerLayer.get_layer_monitoring_channels
Problem is that when raw_space in this function is composite space:
raw_space CompositeSpace(Conv2DSpace(shape=(4, 4), num_channels=64, axes=('c', 0, 1, 'b'), dtype=float32), Conv2DSpace(shape=(4, 4), num_channels=64, axes=('c', 0, 1, 'b'), dtype=float32), Conv2DSpace(shape=(4, 4), num_channels=64, axes=('c', 0, 1, 'b'), dtype=float32), Conv2DSpace(shape=(4, 4), num_channels=64, axes=('c', 0, 1, 'b'), dtype=float32), Conv2DSpace(shape=(4, 4), num_channels=64, axes=('c', 0, 1, 'b'), dtype=float32))
and raw_state:
raw_state (Reshape{2}.0, Reshape{2}.0, Reshape{2}.0, Reshape{2}.0, Reshape{2}.0)
and inside function when you execute code line:
raw_space.validate(raw_state)
you get error:
ValueError: The value of a Conv2DSpace batch must be 4D, got 2 dimensions for Reshape{2}.0.
Bug started with commit: cb6b2510af7f607f729822f8bddc54cace5864cf

JesseLivezey:
This is motivated by the following problem:

FlattenerLayers interact with their raw_layers when costs are called from them and when monitoring channels are requested. Both of these things generally do not want the FlattenerLayer output, but rather the output of the layers in the CompositeLayer the FlattenerLayer wraps. This can't be done with another format_as since this adds more theano ops and then the costs and monitoring channels can't get the actual results of fprop easily.

Fixes problems referenced here:
https://github.com/lisa-lab/pylearn2/issues/1301

and might help:
https://github.com/lisa-lab/pylearn2/pull/1290

- [x] confirm that this fixes known problems (did this locally)
 - [x] add tests to pylearn2
- [x] write tests to confirm that this works for all relevant spaces
 - [x] write tests for symbolic undo
 - [x] write tests for numeric undo
 - [x] Check dtype conversion
- [x] remove duplicate code with _undo_cast, etc.

### DIFF
--- a/pylearn2/space/__init__.py
+++ b/pylearn2/space/__init__.py
@@ -224,6 +224,34 @@ def _cast(arg, dtype):
         raise TypeError("Unsupported arg type '%s'" % str(type(arg)))
 
 
+def _undo_op(arg, string, strict=False):
+    """
+    Undo symbolic op if string is in str(op).
+
+    Returns <arg> untouched if there was no symbolic op.
+
+    Parameters
+    ----------
+    arg : any symbolic variable.
+    string : str
+        String that specifies op.
+    strict : bool
+        Whether to force op undo or not (default False).
+    """
+
+    if hasattr(arg.owner, 'op'):
+        owner = arg.owner
+        if string in str(owner.op):
+            return owner.inputs[0]
+        elif strict:
+            raise ValueError(string + ' not found in op ' +
+                             str(owner.op) + '.')
+    elif strict:
+        raise ValueError(string + ' op not found in variable ' +
+                         str(arg) + '.')
+    return arg
+
+
 class Space(object):
     """
     A vector space that can be transformed by a linear operator.
@@ -546,6 +574,88 @@ class Space(object):
         """
 
         raise NotImplementedError("%s does not implement _format_as_impl()." %
+                                  type(self))
+
+    def undo_np_format_as(self, batch, space):
+        """
+        Returns a numeric batch (e.g. a numpy.ndarray or scipy.sparse sparse
+        array), with formatting from space undone.
+
+        This is just a wrapper around self._undo_format_as, with an extra check
+        to throw an exception if <batch> is symbolic.
+
+        Parameters
+        ----------
+        batch : numpy.ndarray, or one of the scipy.sparse matrices.
+            Array which lies in this space.
+        space : Space
+            Space to undo formatting from.
+
+        Returns
+        -------
+        numpy.ndarray or one of the scipy.sparse matrices
+            The formatted batch.
+        """
+
+        self._check_is_numeric(batch)
+
+        return space.np_format_as(batch=batch,
+                                  space=self)
+
+    def undo_format_as(self, batch, space):
+        """
+        Returns a symbolic batch (e.g. a theano.tensor or theano.sparse
+        array), with formatting from space undone.
+
+        This is just a wrapper around self._undo_format_as, with an extra check
+        to throw an exception if <batch> is symbolic. Formatting to space
+
+        Parameters
+        ----------
+        batch : numpy.ndarray, or one of the scipy.sparse matrices.
+            Array which lies in this space.
+        space : Space
+            Space to undo formatting from.
+
+        Returns
+        -------
+        A symbolic Theano variable
+            The batch formatted as self.
+        """
+        self._check_is_symbolic(batch)
+        space.validate(batch)
+        self._check_sizes(space)
+
+        batch = self._undo_format_as_impl(batch=batch,
+                                          space=space)
+        # Checks if batch belongs to this space
+        self.validate(batch)
+
+        return batch
+
+    def _undo_format_as_impl(self, batch, target_space):
+        """
+        Actual implementation of undo_format_as.
+        Undoes target_space_formatting.
+        Note that undo_np_format_as calls np_format_as.
+
+        Parameters
+        ----------
+        batch : a theano symbol, or a nested tuple thereof
+            Implementations of this method may assume that batch lies in
+            space (i.e. that it passed self._validate(batch) without throwing
+            an exception).
+        target_space : A Space subclass
+            The space to undo batch formatting from.
+
+        Returns
+        -------
+        A symbolic Theano variable
+            The batch, converted from target_space, back to self.
+        """
+
+        raise NotImplementedError("%s does not implement "
+                                  "_undo_format_as_impl()." %
                                   type(self))
 
     def validate(self, batch):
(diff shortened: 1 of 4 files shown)


JesseLivezey:
@lamblin, this PR should help with some of the FlattenerLayer problems that have come up recently. I basically wrote ```space.undo_format_as()``` by undoing each symbolic op from ```space.format_as()```.

I wrote a bunch of tests to check direct space conversions. I can also add a test network that was broken before--a FlattenerLayer that wraps ConvLayers.

pylearn2/space/__init__.py
```diff
@@ -548,6 +548,129 @@ def _format_as_impl(self, is_numeric, batch, target_space):
...
+        -------
+        WRITEME
+            The formatted batch
+        """
+
+        self._check_is_numeric(batch)
+
+        return self._undo_format_as(is_numeric=True,
+                                    batch=batch,
+                                    space=space)
```
lamblin: I'm not sure why delegate the work to _undo_format_as, since we could call np_format_as here directly.

pylearn2/space/__init__.py
```diff
@@ -548,6 +548,129 @@ def _format_as_impl(self, is_numeric, batch, target_space):
...
+        Returns
+        -------
+        WRITEME
+            The formatted batch
+        """
+        self._check_is_symbolic(batch)
+
+        return self._undo_format_as(is_numeric=False,
+                                    batch=batch,
+                                    space=space)
```
lamblin: Same thing here, what is done by _undo_format_as(is_numeric=False, ...) could simply be put here.

JesseLivezey:
@lamblin, I was mostly just following the structure of format_as for no good reason. I've removed the unnecessary methods.

I also cleaned up some of the duplicated code with an _undo_op function.

lamblin:
Makes a lot of sense to me.
Could you please rebase on the latest master, and remove the merge commit? If you use `git rebase -i`, the merge commit may automatically disappear, or you could explicitly remove it.
Then you would have to force the `git push` to this branch.

JesseLivezey:
@lamblin, I've rebased onto master, but there is a failure coming from a ```float < str``` comparison when the monitoring channels are being sorted (I think). It looks like it is only happening in the python 3 tests.

dwf:
Build failing. Is this still WIP? Can you add [WIP] to the title?

JesseLivezey:
@dwf, it seems to be failing because of a python 3 problem that I don't think I've caused. I described in in my previous comment.

lamblin:
Trying to trigger merge with the latest master, which may have fixed the python3 issue.

pylearn2/space/__init__.py
```diff
@@ -548,6 +576,87 @@ def _format_as_impl(self, is_numeric, batch, target_space):
...
+        Parameters
+        ----------
+        batch : numpy.ndarray, or one of the scipy.sparse matrices.
+            Array which lies in this space.
+        space : Space
+            Space to undo formatting from.
+
+        Returns
+        -------
+        WRITEME
```
lamblin: You can simply put the type here, for instance `numpy.ndarray, or one of the scipy.sparse matrices` instead of WRITEME. You do not have to name the returned value.

pylearn2/space/__init__.py
```diff
@@ -548,6 +576,87 @@ def _format_as_impl(self, is_numeric, batch, target_space):
...
```
lamblin: idem.

pylearn2/space/__init__.py
```diff
@@ -548,6 +576,87 @@ def _format_as_impl(self, is_numeric, batch, target_space):
...
+
+    def _undo_format_as_impl(self, batch, target_space):
+        """
+        Actual implementation of undo_format_as.
+        Undoes target_space_formatting.
+        Note that undo_np_format_as calls np_format_as.
+
+        Parameters
+        ----------
+        batch : atheano symbol, or a nested tuple thereof
```
lamblin: " a theano"

pylearn2/space/__init__.py
```diff
@@ -548,6 +576,87 @@ def _format_as_impl(self, is_numeric, batch, target_space):
...
+            Implementations of this method may assume that batch lies in
+            space (i.e. that it passed self._validate(batch) without throwing
+            an exception).
+        target_space : A Space subclass
+            The space to undo batch formatting from.
+
+        Returns
+        -------
+        The batch, converted from target_space, back to self.
```
lamblin: First line has to be a type

pylearn2/space/__init__.py
```diff
@@ -1998,7 +2235,8 @@ def recursive_shared(batch):
     @functools.wraps(Space._format_as_impl)
     def _format_as_impl(self, is_numeric, batch, space):
         """
-        Supports formatting to a single VectorSpace, or to a CompositeSpace.
+        Supports undo formatting to a single VectorSpace,or to a
```
lamblin: This change should probably be reverted, this is the docstring of `_format_as_impl`, not `_undo_format_as_impl`

pylearn2/space/__init__.py
```diff
@@ -2084,6 +2322,105 @@ def recursive_format_as(orig_space, batch, dest_space):
                                   " does not know how to format as " +
                                   str(space))
 
+    @functools.wraps(Space._undo_format_as_impl)
+    def _undo_format_as_impl(self, batch, space):
+        """
+        Supports formatting to a single VectorSpace, or to a CompositeSpace.
```
lamblin: And the other way around here.

pylearn2/space/__init__.py
```diff
@@ -2321,4 +2658,3 @@ def _batch_size_impl(self, is_numeric, batch):
         # There is no way to know how many examples would actually
         # have been in the batch, since it is empty. We return 0.
         self._validate(is_numeric, batch)
-        return 0
```
lamblin: Why remove that line? the comment says "We return 0."

lamblin:
The buildbot is passing now.

pylearn2/space/__init__.py
JesseLivezey: Fixes these in new commit.

pylearn2/space/__init__.py
JesseLivezey: Changed these back and make "undo" version more clear.

pylearn2/space/__init__.py
JesseLivezey: This wasn't intentional, fixed in new commit.

JesseLivezey:
@lamblin, I've made the changes you've requested.

lamblin:
Great, thanks!

Issue #1319, opened 2014-12-20
Adult training and test datasets are incompatible as written


fishcorn:
The root of the problem is the fact that there is one more country category in adult.data than there is in adult.test. This results in one more dimension in the 'train' set than in the 'test' set. If used together this is a large problem, because incompatibly encoded datasets can't be reliably tested against each other.

Reproducing the problem:
Download adult.data and adult.test from https://archive.ics.uci.edu/ml/machine-learning-databases/adult/ and run:

```
grep Holand-Netherlands adult.{data,test}
```

Only adult.data has this category.

In a python shell:

```
In [8]: print adult.adult('train').X.shape
(30162, 104)
In [9]: print adult.adult('test').X.shape
(15060, 103)
```

This results in incompatible dimensions when running 'test' as a monitoring dataset.

Also problematic, even if one were to "hide" the problem by simply deleting the record containing "Holand-Netherlands," is that the code to load the adult dataset appears to use the following construct:

```
unique_values = list(set(data))
```

I assume this is here to get a unique subset of the categories, but this won't sort the strings, at least in python 2.7, which is what I use:

```
In [12]: list(set(["alpha","omega","epsilon"]))
Out[12]: ['alpha', 'omega', 'epsilon']
```

So it's still possible to get a permuted encoding between 'train' and 'test'. Even with sorting, this doesn't account for the missing category.

I can envision two possible solutions to this:
1. Parse "adult.names" for the categories and use those lists as the source for the encodings
2. Encode 'test' against the categories from 'train'.

I used (1) for my tests but it has the drawback of breaking everything built against the current code. The second means that you have to load 'train' every time you load 'test' but it won't break anything using 'train'.

lamblin:
I don't think anything currently relies on the current code, especially because it is broken.
Using the lists in `adult.names` make sense, but parsing it may be a hassle, especially if the format changes in the future (it already has, as there is an `old.adult.names`).
So maybe we should explicitly put the list of labels in `adult.py`, in the same order, and use that.
This will avoid the need to read `adult.test` or `adult.train` every time.
Would you be willing to submit a PR?
Thanks a lot.

fishcorn:
That makes more sense. Ok, I'll proceed with the assumption that I can make
a breaking change to the dataset, and produce a PR based on that. It
probably won't be immediate, but when I can get to it.

lamblin:
Thanks!

PR #1339, opened 2015-01-11
Added more test codes from David Kruger, dealing with cosdataset now.

hantek:
Seems that David Kruger has already finished dealing with the following 5 datasets tests in ticket https://github.com/lisa-lab/pylearn2/pull/666:
 * iris
 * ocr
 * hepatitis
 * avicenna
 * cifar100
So here we just copy his codes into this PR and check if that will work first.

We'll find a way to deal with cosdataset in future commits of this PR.

### DIFF
--- a/pylearn2/datasets/tests/test_adult.py
+++ b/pylearn2/datasets/tests/test_adult.py
@@ -1,38 +1,23 @@
 """
-Testing class that simply checks to see if the adult dataset
+Test code for adult.py
 """
-import unittest
-import numpy as np
+import numpy
 from pylearn2.datasets.adult import adult
 from pylearn2.testing.skip import skip_if_no_data
 
-
-class TestAdult(unittest.TestCase):
+def test_adult():
     """
-    Parameters
-    ----------
-    None
-
-    Notes
-    -----
-    Testing class that simply checks to see if the adult dataset
-    is loadable
+    Tests if it will work correctly for train and test set.
     """
-    def setUp(self):
-        """
-        Skips test if data does not exist
-        """
-        skip_if_no_data()
+    skip_if_no_data()
+    adult_train = adult(which_set='train')
+    assert (adult_train.X >= 0.).all()
+    assert adult_train.y.dtype == bool
+    assert adult_train.X.shape == (30162, 104)
+    assert adult_train.y.shape == (30162, 1)
 
-    def test_adult():
-        """
-        Tests that the dataset loads correctly
-        and that there is no inf in the data
-        """
-        skip_if_no_data()
-        data = adult(which_set='train')
-        assert data.X is not None
-        assert data.X is not np.inf
-        data = adult(which_set='test')
-        assert data.X is not None
-        assert data.X is not np.inf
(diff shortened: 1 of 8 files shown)


memimo:
Travis is failing because test_cos_dataset.py has not docstring.
Please also add more tests for avicena and adult datasets, at least test that they don't have nan values, etc.

hantek:
Hi Mehdi, 

I have updated some tests according to our discussion.

memimo:
Seems like you need a rebase.

hantek:
Hi Mehdi,

I think this time it is ready to merge : )

memimo:
Commits should be representative of logical division of the work not work progress. So when you adding a new fix like pep8 instead of making a new commit you can update last commit with git --amend.
But now that you have made a new commits, you can clear history using interactive git rebase. Here is a short tutorial in how to do so: http://git-scm.com/book/en/v2/Git-Tools-Rewriting-History
You probably need to git push -f after rebase.
It would be ready for merge after git rebase with cleaner commit history. Thanks!

hantek:
So I type the following commit, is that right? 

git rebase HEAD^^
git push -f

Asking before I leap in case I ruin the branch...

dwf:
@memimo I think you mean "git commit --amend"

hantek:
I find another way, but it is not using rebase:

git checkout test^
"do the modifications in the second PEP8 respect commit again"
git add
git commit --amend
git push -f

Is that right?

dwf:
That's going to blow away the most recent commit on the test branch, but if
that's acceptable (as you're doing all the modifications again) sure.

hantek:
Thanks @dwf seems it works. Still wanna to know the rebase way in this case. Thanks for helping me!

dwf:
Basically rebase -i, whatever you have as $EDITOR will pop up a file for
you to edit, you can set some commits to "s" (for "squash") into previous
commit.

pylearn2/datasets/tests/test_adult.py
```diff
@@ -1,38 +1,24 @@
...
-        assert data.X is not None
-        assert data.X is not np.inf
-        data = adult(which_set='test')
-        assert data.X is not None
-        assert data.X is not np.inf
+    adult_test = adult(which_set='test')
+    assert (adult_test.X >= 0.).all()
+    assert adult_test.y.dtype == bool
+    assert adult_test.X.shape == (15060, 103)
+    assert adult_test.y.shape == (15060, 1)
```
lamblin: This code asserts that the shape of the training and test sets are different.
This will have to change in the future, as for the moment it means they are not compatible! See #1319

lamblin:
The file core.6859 should never have been added!
@memimo, @hantek, can one of you remove it?

memimo:
oh, I didn't notice that file. I'm gonna revert this PR. @hantek please make a new PR without that file.

hantek:
@lamblin Yes, and the new PR is here: https://github.com/lisa-lab/pylearn2/pull/1365

PR #1340, opened 2015-01-12
Added shell script to download the SVHN dataset

sandeepkbhat:
[chg] Added ./pylearn2/scripts/datasets/download_svhn.sh script for SVHN dataset download to the $PYLEARN2_DATA_PATH/SVHN/format2 directory. The script is closely based on download_cifar10.sh, but it uses a verbose wget or curl.


lamblin:
Thanks!

PR #1342, opened 2015-01-12
EpochMonitor callback for monitoring progress within epochs

AdamStone:
When using a large training set, it can take a long time for the first epoch to complete and display the first monitor update. I wanted to see that progress was being made during the epoch, so I wrote an SGD callback to do this using the method suggested by Ian here https://groups.google.com/forum/#!topic/pylearn-users/Mz1_1yqSYeA

This can be used to print simple one-line updates giving the current epoch and batch at a specified rate, and/or to print monitor updates within epochs at a specified rate. 

I also made the file consistent with PEP8.

### DIFF
--- a/pylearn2/devtools/tests/test_format.py
+++ b/pylearn2/devtools/tests/test_format.py
@@ -131,7 +131,6 @@ whitelist_pep8 = [
     "training_algorithms/tests/test_default.py",
     "training_algorithms/default.py",
     "training_algorithms/training_algorithm.py",
-    "training_algorithms/sgd.py",
     "distributions/tests/test_mnd.py",
     "distributions/parzen.py",
     "distributions/uniform_hypersphere.py",
--- a/pylearn2/training_algorithms/sgd.py
+++ b/pylearn2/training_algorithms/sgd.py
@@ -26,8 +26,8 @@ from pylearn2.space import CompositeSpace, NullSpace
 from pylearn2.train_extensions import TrainExtension
 from pylearn2.training_algorithms.training_algorithm import TrainingAlgorithm
 from pylearn2.training_algorithms.learning_rule import Momentum
-from pylearn2.training_algorithms.learning_rule import MomentumAdjustor \
-        as LRMomentumAdjustor
+from pylearn2.training_algorithms.learning_rule import (
+    MomentumAdjustor as LRMomentumAdjustor)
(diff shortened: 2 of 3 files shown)


dwf:
Could you rebase the commits you've added against current master? That would make it easier to review this. But it sounds really useful, so thank you in advance.

AdamStone:
Sure, I suppose it's about time I figure out how rebasing works anyway

AdamStone:
I think I did it right, if not let me know.

lamblin:
It looks good, thanks!
Now that you have fixed all sort of pep8 violations in that file, could you take it off the pep8 whitelist in https://github.com/lisa-lab/pylearn2/blob/master/pylearn2/devtools/tests/test_format.py#L145 ?
Also, don't worry about it now, but for the future, if you can separate commits that do only reformatting from commits that change the code itself, it can make things easier to review.
@dwf, do you want to take another look?

AdamStone:
Sure, that sounds like a good idea. I was originally just intending to add the EpochMonitor code, and I fixed the PEP8 without really thinking about it in order to pass the automated tests.

AdamStone:
Hmm.. Somehow the build is now failing, with PEP8 problems that apparently aren't recognized by my check PEP8 script. I can fix them, but I wonder why it passed the tests before.

lamblin:
Sorry for the reviewing delay.
Do you think you could add a small test, based on your YAML example, so that this feature does not break in the future? In particular, you could check that the right number of values are in the monitoring channels.
You can use the classes (datasets, and so on) defined in `pylearn2/testing` for dummy objects.
@dwf did you have other comments?

AdamStone:
Sure, I'll give it a try.

AdamStone:
It seems I'm failing my own test in the build check, but I'm not seeing why. If I run the test myself, there's no assertion error. Maybe I'm missing something about how the testing is supposed to work?

lamblin:
The buildbot launches the tests in different settings, including with "floatX=float32" in the theano flags, and using python 3.
My guess is that `n_batches/monitor_rate` is interpreted as floating point division on python 3, and integer division in python 2, so they could give different results. `n_batches // monitor_rate` should be integer division for both.

AdamStone:
Ah yes, that would explain the discrepancy. I didn't see 'from __future__ import division', so I thought I was safe.

AdamStone:
Test passing now.

PR #1345, opened 2015-01-14
Load HDF5 dataset with 1D label vector

fvisin:
I have a couple of doubts: 
1) Is it ok to load every vector as a one-column matrix? Is there a case in which we don't want this?
2) Should all this be handled in the "format_as" method of IndexSpace instead of changing the way we load the data? In this case, should we modify also the other spaces?

### DIFF
--- a/pylearn2/datasets/dense_design_matrix.py
+++ b/pylearn2/datasets/dense_design_matrix.py
@@ -790,7 +790,7 @@ class DenseDesignMatrix(Dataset):
 
             # This is to support old pickled models as well to trigger the
             # conversion of the labels vector into a 2D one-hot matrix if
-            # y_labels is specified
+            # y_labels is specified and the cost requires it
             if getattr(self, 'y_labels', None) is not None:
                 y_space = IndexSpace(dim=dim, max_labels=self.y_labels)
             elif getattr(self, 'max_labels', None) is not None:
--- a/pylearn2/datasets/dense_design_matrix.py
+++ b/pylearn2/datasets/dense_design_matrix.py
@@ -129,7 +129,8 @@ class DenseDesignMatrix(Dataset):
         If y contains labels then y_labels must be passed to indicate the
         total number of possible labels e.g. 10 for the MNIST dataset
         where the targets are numbers. This will make the set use
-        IndexSpace.
+        IndexSpace that will convert the label data in a one-hot matrix if
+        needed.
 
     See Also
     --------
@@ -786,7 +787,10 @@ class DenseDesignMatrix(Dataset):
                 dim = 1
             else:
                 dim = self.y.shape[-1]
-            # This is to support old pickled models
+
+            # This is to support old pickled models as well to trigger the
+            # conversion of the labels vector into a 2D one-hot matrix if
+            # y_labels is specified
             if getattr(self, 'y_labels', None) is not None:
                 y_space = IndexSpace(dim=dim, max_labels=self.y_labels)
             elif getattr(self, 'max_labels', None) is not None:
--- a/pylearn2/datasets/hdf5.py
+++ b/pylearn2/datasets/hdf5.py
@@ -147,7 +147,6 @@ class HDF5Dataset(DenseDesignMatrix):
         * HDF5ViewConverter is used instead of DefaultViewConverter
         * Data specs are derived from topo_view, not X
         * NaN checks have been moved to HDF5DatasetIterator.next
-        * Support for "old pickled models" is dropped.
 
         Note that y may be loaded into memory for reshaping if y.ndim != 2.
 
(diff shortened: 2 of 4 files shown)


pylearn2/datasets/dense_design_matrix.py
```diff
@@ -786,7 +787,10 @@ def set_topological_view(self, V, axes=('b', 0, 1, 'c')):
                 dim = 1
             else:
                 dim = self.y.shape[-1]
-            # This is to support old pickled models
+
+            # This is to support old pickled models as well to trigger the
+            # conversion of the labels vector into a 2D one-hot matrix if
+            # y_labels is specified
```
lamblin: That's not exactly True. The conversion to one-hot will not be done unless the cost requires it. Some costs work with IndexSpace.

pylearn2/datasets/hdf5.py
```diff
@@ -233,7 +239,26 @@ def next(self):
...
+                # example test_hdf5_topo_view(), where where i
+                # next.index.shape = (10,) and data is 'data': <HDF5
+                # dataset "y": shape (10, 3), type "<f8">
+                # I think it would be better to explicitly check if
+                # next_index.shape is incompatible with data.shape, for
+                # instance checking if next_index.ndim == data.ndim
+                if data.ndim > 1:
+                    this_data = data[next_index, :]
+                else:
+                    raise
```
lamblin: I don't understand how that would work.
I cannot imagine a case where `data[next_index]` would be a problem, but `data[next_index, :]` would be OK.

pylearn2/datasets/hdf5.py
```diff
@@ -233,7 +239,26 @@ def next(self):
...
+            # Check if the dataset data is a vector and transform it into a
+            # one-column matrix. This is needed to automatically convert the
+            # shape of the data later (in the format_as method of the
+            # Space.)
+            if this_data.ndim == 1:
+                this_data = this_data[:, np.newaxis]
```
lamblin: This is too general. Some spaces may require data of dimensionality 1 in the future, and this would not work.

lamblin:
So, to answer your questions:
> 1) Is it ok to load every vector as a one-column matrix? Is there a case in which we don't want this?
There is no current case where it would be a problem, but it could become one in the future. It is also confusing.

> 2) Should all this be handled in the "format_as" method of IndexSpace instead of changing the way we load the data? In this case, should we modify also the other spaces?
In theory, we could make a vector of ints a valid value for an IndexSpace. The problem would be that there would not be a canonical representation of that value in an IndexSpace: if that vector was converted to another space and back, it may come back as a column vector, which would be different from the original.

I really think the answer is:
- clarify that y has to be a 2D tensor
- fix the original dataset so that it is the case.

pylearn2/datasets/dense_design_matrix.py
fvisin: Ok, I just thought that it might be useful to know that it is not only to support old models, to avoid a "support dropped" situation as in the hdf5 equivalent.

pylearn2/datasets/hdf5.py
fvisin: This code is not mine, I just added a comment based on what I experienced debugging these lines. As I wrote, if you execute test_hdf5_topo_view() in test_hdf5.py data[next_index] will trigger an exception while data[next_index, :] works.

fvisin:
> There is no current case where it would be a problem, but it could become one in the future. It is also confusing.

Ok, I agree

> In theory, we could make a vector of ints a valid value for an IndexSpace. The problem would be that there would not be a canonical representation of that value in an IndexSpace: if that vector was converted to another space and back, it may come back as a column vector, which would be different from the original.

Do you think it would make sense to check if it is 'y' and 'y_labels' is set and convert it only in that case?

fvisin:
Ok, I deleted my workaround and modified the error message to be more informative.
I am leaving the code to set y_space to an IndexSpace if y_labels or max_labels are specified.

pylearn2/datasets/hdf5.py
lamblin: OK, I just tried it with h5py.
The problem is that, when `next_index` is an `ndarray` and not a list, `data[next_index]` is interpreted as indexing by a mask, not by a list of indices, and it fails with the following error:
```
TypeError: PointSelection __getitem__ only works with bool arrays
```
The solution is indeed to force the interpretation of `next_index` as an index in only one dimension, but a more robust way of doing that is to always call `data[next_index, ...]`, which works even if `data` has only one dimension.
Could you update the comment again, remove the try/catch and replace it by `data[next_index, ...]`?
If my comment is not clear enough, let me know and I'll submit a PR to yours.

pylearn2/utils/iteration.py
```diff
@@ -816,11 +816,21 @@ def __init__(self, dataset, subset_iterator, data_specs=None,
...
+                                "convert it into a one-column 2D matrix i.e. "
+                                "its shape should be [num_samples, 1]. You "
+                                "should also specify the y_labels parameter "
+                                "so that pylearn2 will use an IndexSpace "
+                                "instead of a VectorSpace. This will ensure "
+                                "an automatic conversion of the vector into a "
+                                "one-hot matrix when needed"
+                            )
+                            # raise, preserving also the original message
                             reraise_as(ValueError(msg))
```
lamblin: It is nice to have a more precise error messge, but please keep the original message (`e`), and do not add the new message in all cases: there are many other reasons a ValueError could be raised at that point.
For instance, the index for a class could be out of bound, the user could try to convert an RGB image into grayscale, etc. In those cases, the new error message would not make sense.

Actually, I think the best place to put such an error message would be in IndexSpace itself, in the `np_validate` method: there, you know exactly which conversion is attempted.

fvisin:
I close this PR cause I wrote a completely new hdf5 handler: https://github.com/lisa-lab/pylearn2/pull/1375/files

The error message improvement part of this PR is still valid, but I will make a different PR for that.

PR #1347, opened 2015-01-15
fix Reshape error and "Input dimension mis-match" when using Softmax layer with binary_target_dim > 1

minhlab:
Using `binary_target_dim` rather than 1 will result in these errors:

````
ValueError: Input dimension mis-match. (input[0].shape[0] = 1968, input[2].shape[0] = 123)
Apply node that caused the error: Elemwise{Composite{[add(i0, mul(i1, i2))]}}[(0, 2)](Flatten{1}.0, TensorConstant{(1,) of 60000}, ARange.0)
Inputs types: [TensorType(int64, vector), TensorType(int64, (True,)), TensorType(int64, vector)]
Inputs shapes: [(1968,), (1,), (123,)]
Inputs strides: [(8,), (8,), (8,)]
Inputs scalar values: ['not scalar', array([60000]), 'not scalar']
````

and

````
ValueError: total size of new array must be unchanged
Apply node that caused the error: Reshape{1}(monitoring_targets, MakeVector.0)
Inputs types: [TensorType(int64, matrix), TensorType(int64, vector)]
Inputs shapes: [(10, 16), (1,)]
Inputs strides: [(136, 8), (8,)]
Inputs scalar values: ['not scalar', array([10])]
````

This commit fixes them.

### DIFF
--- a/pylearn2/models/mlp.py
+++ b/pylearn2/models/mlp.py
@@ -1432,7 +1432,8 @@ class Softmax(Layer):
             flat_Y.name = 'flat_Y'
             flat_log_prob = log_prob.flatten()
             flat_log_prob.name = 'flat_log_prob'
-            range_ = T.tile(T.arange(Y.shape[0]), (self.binary_target_dim,))
+            range_ = T.tile(T.arange(Y.shape[0]).dimshuffle(0,'x'), 
+                            (1,self.binary_target_dim)).flatten()
             flat_indices = flat_Y + range_ * self.n_classes
             flat_indices.name = 'flat_indices'
             log_prob_of = flat_log_prob[flat_indices].dimshuffle(0, 'x')
--- a/pylearn2/models/mlp.py
+++ b/pylearn2/models/mlp.py
@@ -1433,10 +1433,10 @@ class Softmax(Layer):
             flat_log_prob = log_prob.flatten()
             flat_log_prob.name = 'flat_log_prob'
             range_ = T.tile(T.arange(Y.shape[0]).dimshuffle(0,'x'), 
-                            (1,self.binary_target_dim)).flatten()
+                            (1, self.binary_target_dim)).flatten()
             flat_indices = flat_Y + range_ * self.n_classes
             flat_indices.name = 'flat_indices'
-            log_prob_of = flat_log_prob[flat_indices].dimshuffle(0, 'x')
+            log_prob_of = flat_log_prob[flat_indices].reshape(Y.shape, ndim=2)
             log_prob_of.name = 'log_prob_of'
 
         else:
--- a/pylearn2/models/mlp.py
+++ b/pylearn2/models/mlp.py
@@ -1137,7 +1137,9 @@ class Softmax(Layer):
         number of targets here so that an IndexSpace of the proper dimension
         can be used as the target space. This allows the softmax to compute
         the cost much more quickly than if it needs to convert the targets
-        into a VectorSpace.
+        into a VectorSpace. With binary_target_dim>1, you can use one layer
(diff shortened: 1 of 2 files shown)


lamblin:
Do you have an example that triggers the error, so we can check that the changes have the expected result?
Making it a unittest will ensure that it continues working in the future.

minhlab:
I was training this model when I got the errors:

````
model: !obj:pylearn2.models.mlp.MLP {
        layers: [
            !obj:pylearn2.sandbox.nlp.models.mlp.ProjectionLayer {
                layer_name: 'projection',
                dim: 50,
                irange: 0.01,
            }, !obj:pylearn2.models.mlp.Tanh {
                layer_name: 'tanh',
                dim: 10,
                irange: 0.01
            }, !obj:pylearn2.sandbox.nlp.models.mlp.Softmax {
                n_classes: *max_labels,
                layer_name: 'softmax',
                irange: 0.01,
                binary_target_dim: 16 # window*2
            }
        ],
        input_space: !obj:pylearn2.space.IndexSpace {
            dim: 1,
            max_labels: *max_labels
        }
    },
````

I added a test but didn't run it through... Might need modification...

minhlab:
Bugs solved. Unit test added. I'm sorry for the many small commits. Next time I'll use pep8 and test things more carefully before committing.

lamblin:
Thanks for the update.
I'm still trying to understand the logic of what the Softmax layer should do with binary_target_dim > 1.
Currently, there is only one softmax, with `n_classes` inputs, being applied on each of the examples in the minibatch.
When more than one binary target is provided, what does that mean? With your changes, it looks like it is trying to maximize the log probability of each of the target indices at the same time, in the same probability vector.
Is it indeed what you want to do in your case? If it is, then it should be explained clearly in the documentation, because it could be quite confusing: users may expect that as many softmax as binary targets are computed for each example, each of the softmax trying to predict the corresponding target. In fact, maybe that should be the default interpretation, even if it is not currently implemented.

minhlab:
I don't know what the original developers had in mind so I implemented it in the way I think the Skip-gram model works (Mikolov et al. 2013; especially [word2vec source code](https://code.google.com/p/word2vec/source/browse/trunk/word2vec.c)). The softmax layer tries to simultaneously predict a **bag** of N words (i.e. order is not important, the same element can be included more than one time). For example, if it predict the word "dog" appears with probability 0.001 and that word turns out to appear twice, the cost will be (-2 * log 0.001). There's a pitfall though. If one uses a VectorSpace for output (not using `binary_target_dim`), repeated items can only be represented once so the output will be equivalent to a set. While using an IndexSpace (with `binary_target_dim`), it's very hard to represent a set. Therefore, the semantics of the two ways will become different, not just speed. Does it make sense to you?

Reference: Mikolov, T., Corrado, G., Chen, K., & Dean, J. (2013). Efficient Estimation of Word Representations in Vector Space, 1–12.

lamblin:
@minhlab That makes perfect sense, yes. Can you add a sentence in the docstring of the parameter `binary_target_dim` of the `Softmax` layer explaining what it means to have it greater than 1 (i.e., that the target is a bag of word, and we try to predict it with only one softmax)?
Regarding the conversion to `VectorSpace`, I'm also wondering if we should use the sum of all one-hot vectors, enabling some of the elements to be 2 or more if they were present more than once in the original `IndexSpace`. But it is a different question...

minhlab:
Added explanation.

lamblin:
Thanks!
Can you rebase on top of the latest master, and push --force to update the PR?

lamblin:
Thanks!

PR #1349, opened 2015-01-15
Naming issue

lmiksys:
Methods and variables in ExamplewiseAddScaleTransform share the same names (_add, _multiply).


lamblin:
Thanks!
Could you add a minimal test case, so that it does not break in the future? Then we can merge the PR.

PR #1353, opened 2015-01-17
Modify projection layer (sandbox) to handle composite space

minhlab:
To be used in multi-task learning.

### DIFF
--- a/pylearn2/sandbox/nlp/models/tests/composite.yaml
+++ b/pylearn2/sandbox/nlp/models/tests/composite.yaml
@@ -38,7 +38,7 @@
     },
     model: !obj:pylearn2.models.mlp.MLP {
         layers: [
-            !obj:models.mlp.ProjectionLayer {
+            !obj:pylearn2.sandbox.nlp.models.mlp.ProjectionLayer {
                 layer_name: 'projection',
                 dim: 4,
                 irange: 0.05,
--- a/pylearn2/sandbox/nlp/models/tests/composite.yaml
+++ b/pylearn2/sandbox/nlp/models/tests/composite.yaml
@@ -1,6 +1,6 @@
 !obj:pylearn2.train.Train {
     dataset: &train !obj:pylearn2.datasets.vector_spaces_dataset.VectorSpacesDataset {
-        data: !obj:pylearn2.sandbox.nlp.models.as_tuple { list_: [
+        data: [
                   !obj:numpy.random.randint { 
                           'low': 0, 
                           'high': &max_labels 10, 
@@ -13,7 +13,7 @@
                         },
                   !obj:numpy.random.normal { 'size':[5,3] },
                   !obj:numpy.random.normal { 'size':[5,5] },
-              ] },
+              ],
         data_specs: [
                 !obj:pylearn2.space.CompositeSpace { components: [
                         &feature1 !obj:pylearn2.space.IndexSpace { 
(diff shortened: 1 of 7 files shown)


pylearn2/sandbox/nlp/models/tests/composite.yaml
```diff
@@ -0,0 +1,78 @@
+!obj:pylearn2.train.Train {
+    dataset: &train !obj:pylearn2.datasets.vector_spaces_dataset.VectorSpacesDataset {
+        data: !obj:pylearn2.sandbox.nlp.models.as_tuple { list_: [
```
lamblin: I think it may be better to fix `VectorSpacesDataset` so it accepts a list as `data`.

lamblin:
At a high level, it makes sense to me.
@bartvm can you have a look as well?

bartvm:
LGTM, although some unit tests that can run on Travis might be a good idea?

pylearn2/sandbox/nlp/models/tests/composite.yaml
minhlab: I think so but hesitate to modify things...

minhlab:
yes, there are

bartvm:
Oops, mistook that for a script, my bad. Looks good then!

pylearn2/sandbox/nlp/models/tests/test_mlp.py
```diff
@@ -0,0 +1,15 @@
+'''
+Created on Jan 17, 2015
+
+@author: Minh Ngoc Le
```
bartvm: You should probably follow the format used everywhere else (e.g. [here](https://github.com/lisa-lab/pylearn2/blob/master/pylearn2/models/model.py#L1-L7)).

minhlab: Should I put `__maintainer__ = "LISA Lab"` and `__email__ = "pylearn-dev@googlegroups"`? I don't know what maintaining this piece of code would be like.

bartvm: Whichever you prefer. No official responsibilities come with putting your name at the top; it's mostly for credit I guess (because for any questions we can find the one who wrote the code anyway through Git blame).

minhlab: OK, I pushed a new version.

pylearn2/sandbox/nlp/models/tests/composite.yaml
minhlab: I mean maybe people didn't accept lists for a reason.

lamblin: In that case, I'm pretty sure there is no particular reason. It may even already work, and the only change needed would be to update the docstring :)
If it does not, then changing the `__init__` method to set `self.data = tuple(data)` should be enough.

minhlab: You're right, I'll do it next week.

minhlab:
I removed `as_tuple` and change things to make it work.

minhlab:
I [merged pylearn2/utils/iteration.py](https://github.com/minhlab/pylearn2/commit/02cb1c7ca7fb366f3868fd172a3458d7209d8c94) with lisa-lab/master because github complained that there was a merge conflict.

lamblin:
I'm afraid the merge commit will confuse the history on github.
Instead of merging in cases of conflicts, we try to _rebase_ on the latest master instead, and force-push on the pull request.
Do you think you can try to do that? Otherwise I'll try to find some time to do it.

minhlab:
I have never had time to learn git carefully but now it starts to make sense to me :D

lamblin:
Thanks!

Issue #291, opened 2013-05-10
Use stuff in pylearn2.datasets.exc to make skippable unit tests for datasets


goodfeli:
In particular, should test FoveatedNORB. It was broken when I wanted to use it this week.

- [x] Avicenna
- [x] cifar100
- [x] CosDatast
- [x] NORBSmall
- [x] FoveatedNorb
- [x] MNISTPlus
- [x] hepatitis
- [x] icml07
- [x] iris
- [x] ocr
- [x] retina
- [x] stl10
- [ ] svhn (Mehdi)

lamblin:
@memimo Can you check the checkboxes for the datasets that have tests?

memimo:
I'm closing this ticket, since SVHN is dependent on the new hdf5 format or possibly fuel.

PR #1355, opened 2015-01-19
Add unit tests

kelvinxu:
@memimo Initialial tests for retina pending comment.

### DIFF
--- a/pylearn2/datasets/tests/test_retina.py
+++ b/pylearn2/datasets/tests/test_retina.py
@@ -1,0 +1,81 @@
+"""
+Tests for ../retina.py
+"""
+import numpy as np
+
+from pylearn2.utils import as_floatX
+from pylearn2.datasets import dense_design_matrix
+from pylearn2.datasets.retina import (foveate_channel,
+                                      defoveate_channel,
+                                      get_encoded_size,
+                                      encode,
+                                      decode,
+                                      RetinaEncodingBlock,
+                                      RetinaDecodingBlock,
+                                      RetinaCodingViewConverter
+                                      )
+
+
+class test_retina():
+    """
+    Parameters
+    ----------
+    None
+
+    Notes
+    -----
+    Testing class that checks retina preprocessing functions
+    """
+
+    def test_transform(self):
+        """
+        This runs tests on the foveate/defoveate channel
+        process for different sizes of input
+        """
+        # square input
+        image = np.random.randn(1, 20, 20)
+        rings = [5, 2]
+        encoded_size = get_encoded_size(image.shape[1], image.shape[2], rings)
+        assert encoded_size == 64, "Different from  previously computed value"
+        self.foveate_defoveate_channel(image, rings, encoded_size)
+
+        # non-square input
+        rec_image = np.random.randn(1, 70, 42)
+        rec_rings = [7, 4, 2]
+        rec_encoded_size = get_encoded_size(rec_image.shape[1],
+                                            rec_image.shape[2], rec_rings)
+        assert encoded_size == 64, "Different from  previously computed value"
+        self.foveate_defoveate_channel(rec_image, rec_rings,
+                                       rec_encoded_size)
+
+    def foveate_defoveate_channel(self, image, rings, encoded_size, bs=1):
+        """
+        Helper function
+
+        Here we foveate the image, which reduces the quality of the rings
+        Defoveate it back into the original image, and see if we
+        once again refoveate whether it returns to the original image
+        """
+        output = np.zeros((bs, encoded_size))
+        foveate_channel(image, rings, output, 0)
+        defoveate_channel(image, rings, output, 0)
+        refoveated_output = np.zeros((bs, encoded_size))
+        foveate_channel(image, rings, refoveated_output, 0)
+        np.testing.assert_allclose(refoveated_output, output)
+
+    def test_encode_decode(self):
+        """
+        Try to encode and decode to get back the same value
+        """
+        topo_X = np.random.randn(1, 20, 25, 5)
+        rings = [5]
+        output = encode(topo_X, rings)
+        reconstruct = decode(output, (20, 25, 5), [5])
+        np.testing.assert_allclose(encode(reconstruct, rings), output)
+
+    def test_RetinaCodingViewConverter(self):
+        """
+        Test to see if the RetinaCodingViewConverter loads using the
+        typically shape in ../norb_small.py
+        """
+        view_converter = RetinaCodingViewConverter((96, 96, 2), (8, 4, 2, 2))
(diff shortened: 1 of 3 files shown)


memimo:
thx, I think, this the right way to go and you can finish the rest.

kelvinxu:
added the note that I mentioned and split out another test.

kelvinxu:
added the note that I mentioned and split out another test.

kelvinxu:
added the note that I mentioned and split out another test.

memimo:
Look good. One last change! Could you put TODO in the note section, so it would be easy later to take another look at it. Will merge after..

kelvinxu:
Added the note section

kelvinxu:
Added the note section

pylearn2/datasets/tests/test_retina.py
```diff
@@ -0,0 +1,79 @@
...
+    -----
+    Testing class that checks retina preprocessing functions
+    """
+    def test_foveate_channel(self):
+        """
+        Foveate_channel implicitly tests downsample_ring and downsample_rect
+        """
+        img = np.array([[[0.51507119, 0.71686988], [0.62864236, 0.75365718]]])
+        output = np.zeros((1, 4))
+        foveate_channel(img, [2, 3], output, 0)
```
lamblin: These values do not make sense: the image should be at least larger than twice the sum of the ring sizes, and the subsampling factors should divide the size of the image being subsampled.
You can check that by calling get_encoded_size, which will raise an exception in that case.
Also, calling get_encoded_size will give you the right shape to give to output.
Also try with non-squared images, and with random values, if possible.

pylearn2/datasets/tests/test_retina.py
```diff
@@ -0,0 +1,79 @@
...
+        """
+        assert get_encoded_size(4, 12, [2]) == 12
+
+    def test_encode(self):
+        """
+        Test encode method
+        """
+        test_input = np.array([[[[0, 0], [1, 1]]]])
+        np.testing.assert_array_equal(encode(test_input, [1]),
+                                      np.array([[0., 1., 0., 1.]]))
```
lamblin: The input has to be (batch, rows, cols, chans), so here it means that the image size is (1, 2), which is too small for a ring of 1, I think.
To make sure these make sense, you could use the same shape as for foveated norb: width=height=96, 2 channels, rings are [8, 4, 2, 2].

pylearn2/datasets/tests/test_retina.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def test_decode(self):
+        """
+        .. note::
+            It is ambiguous what this function is suppose to do based
+            on the documentation, so it is missing a test for now.
+            Testing it would also test defoveate_channel/restore_ring/
+            restore rec.
```
lamblin: What is ambiguous? I think it is supposed to transform the encoded data back in the original format, what else do you think it should do?
However, it may be broken at the moment, I'm having a look.

lamblin: So, it works for 96x96 images, but not for other sizes.
The reason is that there is still a "96" hardcoded in retina.py on line 236, it should be "img_w" instead.

With that fix, you can write a test where you sample a random encoded representation, then decode and encode it, and check that you recover (almost) the original random matrix.

pylearn2/datasets/tests/test_retina.py
```diff
@@ -0,0 +1,79 @@
...
+        """
+        Test to see if the RetinaCodingViewConverter loads
+        """
+        view_converter = RetinaCodingViewConverter((96, 96, 2), (8, 4, 2, 2))
+
+    def test_RetinaDecodingBlock(self):
+        """
+        Test to see if the RetinaDecodingBlock loads
+        """
+        decoding_block = RetinaDecodingBlock((1, 2, 3, 4), (4, 5, 6, 7))
```
lamblin: These numbers do not make any sense together, so if some input validation is added later, it would make the test fail.

kelvinxu:
fixing now.

pylearn2/datasets/tests/test_retina.py
kelvinxu: So there are quite a lot of comments. In fairness, not a lot it was necessarily obvious from the documentation. But I'm glad you pointed it out so it can be done properly.

kelvinxu:
Seems to be passing the tests, let me know if there are more comments.

pylearn2/datasets/tests/test_retina.py
```diff
@@ -0,0 +1,96 @@
...
+        defoveate_channel(image, rings, output, 0)
+        refoveated_output = np.zeros((bs, encoded_size))
+        foveate_channel(image, rings, refoveated_output, 0)
+        np.testing.assert_allclose(refoveated_output, output)
+
+    def test_encode_decode(self):
+        """
+        Try to encode and decode to get back the same value
+        """
+        topo_X = np.random.randn(1, 20, 25, 5)
```
lamblin: Can you change the batch size here, so it gets tested for a value greater than 1?

pylearn2/datasets/tests/test_retina.py
```diff
@@ -0,0 +1,96 @@
...
+        rings = [5]
+        output = encode(topo_X, rings)
+        reconstruct = decode(output, (20, 25, 5), [5])
```
lamblin: Use `rings` as an argument here, instead of copying `[5]`.

pylearn2/datasets/tests/test_retina.py
```diff
@@ -0,0 +1,96 @@
...
+        np.testing.assert_allclose(encode(reconstruct, rings), output)
+
+    def test_RetinaCodingViewConverter(self):
+        """
+        Test to see if the RetinaCodingViewConverter loads using the
+        typically shape in ../norb_small.py
+        """
+        view_converter = RetinaCodingViewConverter((96, 96, 2), (8, 4, 2, 2))
```
lamblin: It looks like the tests are not finished:
- The `RetinaCodingViewConverter` should be tested on data, like the docstring says it does
- both `design_mat_to_topo_view` and `topo_view_to_design_mat` should be tested, and check that they invert each other.

pylearn2/datasets/tests/test_retina.py
kelvinxu: If you put in rings that don't make sense, it will break the test though?

kelvinxu:
all changed.

memimo:
#291

pylearn2/datasets/tests/test_retina.py
lamblin: OK, thanks.

pylearn2/datasets/tests/test_retina.py
lamblin: Can you please test with a batch size other than 1?

lamblin: That had actually been done in `test_transform`.

lamblin:
Great, thanks!

PR #1358, opened 2015-01-19
Asserts to check parameter consistency in WindowAndFlip

fvisin:


lamblin:
Thanks for the PR.
In order to have a cleaner history, without merge and revert, could you rebase and squash all commits into one? The way to do that would be:
- first, create a new branch with your changes (avoid submitting PRs from your `master` branch):
`git checkout -b assert_window_flip` (you can name it however you want, this is just an example)
- then, fetch the latest information about the central repository (it does not update the working files): `git fetch central`, if you named the central lisa-lab/pylearn2 repository `central`, you may have named it `upstream` or something else
- launch a rebase in interactive mode. git will try to re-apply the commits in your branch onto the latest master: `git rebase -i central/master`. This will launch a text editor with a list of commits.
- in that editor, you should have 4 or 5 commits. You can remove the ones you do not want, for instance 272192a and 93dbb98 (which cancel each other out). You can also "squash" commits together, to group all changes into one. When you are finish, save the file and quit the editor.
- if there are conflicts, you can abort the procedure with `git rebase --abort`, or ask me or someone in the lab for help.
- when the rebase procedure is over, you can either submit a new PR, or update the current one by using `git push --force assert_window_flip:master`, to force the update of your remote `master` branch (on github) with your new `assert_window_flip` one.

pylearn2/train_extensions/window_flip.py
```diff
@@ -114,6 +114,15 @@ def __init__(self,
...
         self._flip = flip
 
+        assert isinstance(self._randomize, list), (
+            "The 'randomize' parameter of WindowAndFlip should be a list")
+        assert isinstance(self._randomize_once, list), (
+            "The 'randomize_once' parameter of WindowAndFlip should be a list")
+        assert isinstance(self._center, list), (
+            "The 'center' parameter of WindowAndFlip should be a list")
+        assert isinstance(self._pad_randomized, int), (
+            "The 'pad_randomized' parameter of WindowAndFlip should be an int")
```
lamblin: I think this check should be a bit more robust, and accept numpy int types as well.
You could use `isinstance(self._pad_randomized, pylearn2.utils.py_integer_types)` instead.

fvisin:
Done! 
Thanks for the detailed explanation on how to rebase and squash.
For future reference, the last command of the procedure should be: `git push --force origin assert_window_flip:master`

PR #1361, opened 2015-01-20
Use cudnn max pooling when possible

JesseLivezey:
Equivalent to pylearn2.mlp.max_pool() but with cuDNN pooling.

In a 5 layer mlp with 2 to 3 layers of convolution and pooling, the version with cuDNN was 25-40% faster.

- [x] write a test
- [x] write more thorough comparison test
- [x] fix max pooling when strides > shape

Note that since the implementation of cudnn mean pooling is not the same between versions and theano supports multiple versions, it's not easy to support mean pooling with cudnn and get consistent results. Mabye once cudnn stabilizes it can be added more easily.

### DIFF
--- a/pylearn2/models/mlp.py
+++ b/pylearn2/models/mlp.py
@@ -19,6 +19,7 @@ from theano.compat.six.moves import reduce, xrange
 from theano import config
 from theano.gof.op import get_debug_values
 from theano.sandbox.rng_mrg import MRG_RandomStreams
+from theano.sandbox.cuda.dnn import dnn_available, dnn_pool
 from theano.tensor.signal.downsample import max_pool_2d
 import theano.tensor as T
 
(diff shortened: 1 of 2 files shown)


nouiz:
Line 3487, i think it should be an or.

JesseLivezey:
Fixed 'and' -> 'or'.

the dnn_available() call failed on Travis. It wasn't clear to me why. Should dnn_available() still work on a machine that doesn't have a GPU?

nouiz:
This is a problem with Theano. I did a PR to fix it:

https://github.com/Theano/Theano/pull/2425

When it is merged, we can close and reopen your PR, this will trigger a new
build your PR with the updated Theano.

pylearn2/models/mlp.py
```diff
@@ -3237,9 +3245,16 @@ def fprop(self, state_below):
                                                        "pooling.")
 
             if self.pool_type == 'max':
-                p = max_pool(bc01=d, pool_shape=self.pool_shape,
-                             pool_stride=self.pool_stride,
-                             image_shape=self.detector_space.shape)
+                if dnn_available():
```
nouiz: I think I would move the if inside max_pool. That way, we are sure that the faster version will be use at all place max_pool is used. (and remove duplicate code)

nouiz:
For the tests, I do not know what is needed. Someone more familiar with Pylearn2 will need to answer. But maybe there is already tests that we can just run with and without cudnn.

lamblin:
Regarding the tests, I think a test checking that equivalent calls to the code in `pool_dnn` and `max_pool` give the same numerical value, for a range of input shapes, pooling shape and pooling strides, would be enough.

JesseLivezey:
Does anyone more familiar with theano know if I can turn off the availability of cudnn for part of a test? Currently I had to add an extra argument to max_pool just so I could make sure parts of the test ran with/without cudnn.

nouiz:
This PR will need a rebase to be merged.

To disable cudnn, you can do it by passing a mode to theano.function.

mode=theano.compile.get_default_mode().excluding("cudnn")

dwf:
This needs a rebase but @nouiz, is it ready for merge after that?

JesseLivezey:
I want to make two more changes in addition to the rebase.

1) There is still one thing that I don't understand. I'd like to be able to make theano think that cudnn_available() is false during graph creation (mlp.fprop), not at compilation time (because then max_pool would have already choosed to use cudnn op or regular op). Is this possible? Otherwise, I have to add a flag to mlp.max_pool that only gets used to write the comparison test.

2) I haven't checked, but I'm guessing mean_pool is also broken when pool_stride > pool_shape, which I can also fix if it is.

nouiz:
You can force dnn_available() to return False by plaing in the internal.
But this is brittle.

But I do not understand why you need this, mlp.max_pool already have a
try_dnn parameter. You can use it to trigger the old code path.

For mean_pool, can you make that in a new PR? This way we can merge this
one faster. This one is much more used then mean_pool.

Before merging after the rebase, I would like to just run the tests in
normal mode and in debug mode for all pylearn2 on a computer that support
cudnn and one that do not. Just to be sure we didn't break something. I can
do this after the rebase when Jesse tell me he is ready to merge. I didn't
understood his point 1).

JesseLivezey:
@nouiz, I think this is ready for a final review. You should make sure I'm using SkipTest correctly. I'm also not sure if I need to do anything to force them to run on gpu if possible.

nouiz:
You need to merge my PR to your PR: 

https://github.com/JesseLivezey/pylearn2/pull/1

I had problems with the test and got in wrong direction for fixed. I'm running again the tests with my PR to your PR.

nouiz:
Test passed on my side, when travis tests pass here, we can merge.

PR #1363, opened 2015-01-21
Revert "Added more test codes from David Kruger, dealing with cosdataset...

memimo:
Reverts lisa-lab/pylearn2#1339


lamblin:
Reverting will just make the history uglier. I'll merge #1365 directly.

PR #1368, opened 2015-01-23
Enable Travis container environment

### LINKED ISSUE #1367: Switch to Travis CI container-based environment
Travis's [newer environment](http://docs.travis-ci.com/user/workers/container-based-infrastructure/) works with containers, and jobs start up significantly faster on it (for Blocks total build times went from 35-40 to 20-25 minutes, bartvm/blocks#142). Moreover, if we want to test Pylearn2 with real datasets we could add them to a cache so that they don't need to be redownloaded every run (this is what we do with MNIST for Blocks).

The only downside is that this new infrastructure doesn't allow for `sudo`. That means that either (a) we need to somehow install the binaries without root privileges, which is nearly impossible with `apt-get` and `dpkg`; or (b) we switch to MKL after all (see discussion on #1170). I'd suggest doing the latter.

bartvm:
Small change, but should see significant improvement in Travis CI runtimes. MKL shaves off about half an hour of the total runtime, and on top of that the builds should start sooner.

Fixes #1367.

### DIFF
--- a/.travis.sh
+++ b/.travis.sh
@@ -11,5 +11,5 @@ else
     # download each time. If present run: python make_dataset.py
     cd pylearn2/scripts/tutorials/grbm_smd; { wget -t 1 http://www.iro.umontreal.ca/~lisa/datasets/cifar10_preprocessed_train.pkl && export PKL=true || export PKL=false; }
     cd -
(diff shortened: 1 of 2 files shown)


bartvm:
@lamblin Let me know what you think.

To support my claim: All 5 builds have already started in less than a minute, while normally Python 3 takes a long time to start.

lamblin:
Works for me. Should I merge now, or are more changes needed?

bartvm:
Nope, this should be it, so go ahead.

nouiz:
Could you do the same for Theano? You seem to be able to do it rapidly.

PR #1370, opened 2015-01-26
Add asserts to check input parameter type in window_flip

fvisin:


pylearn2/train_extensions/window_flip.py
```diff
@@ -114,6 +114,15 @@ def __init__(self,
         self._pad_randomized = pad_randomized
         self._flip = flip
 
+        assert isinstance(self._randomize, list), (
+            "The 'randomize' parameter of WindowAndFlip should be a list")
+        assert isinstance(self._randomize_once, list), (
+            "The 'randomize_once' parameter of WindowAndFlip should be a list")
+        assert isinstance(self._center, list), (
+            "The 'center' parameter of WindowAndFlip should be a list")
+        assert isinstance(self._pad_randomized, int), (
```
lamblin: You should use `isinstance(self._pad_randomized, pylearn2.utils.py_integer_types)` here instead.

fvisin: Done.

PR #1373, opened 2015-01-28
import serial in util/__init__.py for simpler importing

capybaralet:
Allows one to access pylearn2/utils/serial.py via "import pylearn2.utils.serial"


goodfeli:
"import pylearn2.utils.serial" already works for me:

>>> import pylearn2.utils.serial
>>>

PR #1375, opened 2015-01-29
Hdf5 file new handler

fvisin:


pylearn2/datasets/hdf5.py
```diff
@@ -2,338 +2,472 @@
...
+            sources = self._get_canonical_sources()
+            spaces = self._get_canonical_spaces()
+            data_specs = (spaces, sources)
+            data = self._get_canonical_data()
+        else:
+            sources = data_specs[1]
+            data = [self.data[s] for s in sources]
+
+        return FiniteDatasetIterator(self,
+                                     mode(data[0].shape[0],
```
fvisin: This doesn't feel right, but apparently it was the default behaviour.
For the future, if we want to be able to iterate over multiple sources I think we should change the interface, e.g. if you want to iterate on _features_ only you should specify _'features'_ instead of _('features', 'targets')_ in _data_specs_. Is there a reason why _target_ is passed along with _features_ but is not used?

pylearn2/datasets/hdf5.py
```diff
@@ -2,338 +2,472 @@
...
+        return_tuple : bool, optional
+            Always return a tuple, even if there is exactly one source
+            of data being returned. Defaults to `False`.
+        """
+        [mode, batch_size, num_batches, rng, data_specs] = self._init_iterator(
+            mode, batch_size, num_batches, rng, data_specs)
+        convert = None
+
+        if data_specs is None:
+            sources = self._get_canonical_sources()
```
fvisin: When data_specs is None and the dataset's data contain ('X', 'y') or ('topo_view', 'y') is it right to pass to the iterator _('features', 'targets')_ and the relative spaces as data_spec?

pylearn2/datasets/hdf5.py
```diff
@@ -2,338 +2,472 @@
...
+        `y` if present. This is needed for compatibility with legacy code but
+        you should not rely on it.
+        """
+        space = []
+        try:
+            space.append(self.spaces['X'])
+        except KeyError:
+            space.append(self.spaces['topo_view'])
+        try:
+            space.append(self.spaces['y'])
```
fvisin: I guess this is a bug, I should return a Composite space in case they are more than one. Is it right?

pylearn2/datasets/dataset.py
```diff
@@ -2,21 +2,25 @@
...
+import warnings
+from pylearn2.utils.iteration import resolve_iterator_class
+
 
 class Dataset(object):
 
     """
     Abstract interface for Datasets.
     """
+    __metaclass__ = abc.ABCMeta
```
lamblin: Is that line supposed to work yet?

pylearn2/datasets/dataset.py
```diff
@@ -73,6 +77,79 @@ def iterator(self, mode=None, batch_size=None, num_batches=None,
...
+        Returns
+        -------
+        iter_obj : object
+            An iterator object implementing the standard Python
+            iterator protocol (i.e. it has an `__iter__` method that
+            return the object itself, and a `next()` method that
+            returns results until it raises `StopIteration`).
+        """
+        # TODO How much of this is possible to put in the Dataset.iterator or
+        # in FiniteDatasetIterator?
```
lamblin: I'm not sure it makes sense to factor it out of DenseDesignMatrix yet, I've always considered the `_iter_*` to be private and used only by DenseDesignMatrix.

pylearn2/datasets/dataset.py
fvisin: Yeah, I was looking at Travis errors right now.
That line is not a problem, the @abstractmethod decorator of iterator though doesn't let you wrap the method from other classes. The test were failing cause DenseDesignMatrix.iterator wrapped Dataset.iterator. I just submitted a commit to fix it, let's see if it will pass the tests.

pylearn2/datasets/dataset.py
```diff
@@ -102,6 +178,8 @@ def get_topo_batch_axis(self):
         Returns the index of the axis that corresponds to different examples
         in a batch when using topological_view.
         """
+        warnings.warn('This method is deprecated and will be unsupported '
+                      'after 27 July 27, 2015')
```
lamblin: I think these warnings should be in the docstring as well, since the `warn` would probably never be called.
It should still be part of DenseDesignMatrix, though, and I think the warning should reflect that, and say it is not going to be part of `Dataset`, but some `DenseDesignMatrix` would still implement it.

pylearn2/datasets/dataset.py
fvisin: I am using it in HDF5.py too, so unless it is wrong to initialize HDF5.iterator in the same way I guess it is better to have it in Dataset rather than having duplicated code.

pylearn2/datasets/hdf5.py
```diff
@@ -2,338 +2,472 @@
...
+            result in a significant speed up. Sensible default values depend
+            on the size of your data and the batch size you wish to use. A
+            rule of thumb is to make a chunk contain 100 - 1000 batches and
+            make sure they encompass complete samples.
+        use_h5py: bool or 'auto', optional
+            Specifies if h5py or pytables should be used. If set to auto
+            pylearn2 will try to use pytables and will switch to h5py if
+            pytables cannot be loaded (e.g. is not installed).
+        kwargs : dict, optional
+            Keyword arguments passed to `DenseDesignMatrix`.
```
lamblin: That docstring should be directly at the class-level, not in the `__init__` method, otherwise it will not get properly displayed online.

pylearn2/datasets/hdf5.py
```diff
@@ -2,338 +2,472 @@
...
+        except KeyError:
+            data.append(self.data['topo_view'])
+        try:
+            data.append(self.data['y'])
+        except KeyError:
+            pass
+        return data
 
-class HDF5DatasetIterator(FiniteDatasetIterator):
+    #@wraps(Dataset.get_data_specs, assigned=(), updated=())
```
fvisin: This seems like a but in the Dataset interface: get_data_specs is used in iteration.755 but is not part of the interface

pylearn2/datasets/hdf5.py
```diff
@@ -2,338 +2,472 @@
...
+            space = tuple([self.spaces[s] for s in source_or_alias])
+            space = CompositeSpace(space)
+        else:
+            space = self.spaces[source_or_alias]
+        return (space, source_or_alias)
 
-    def next(self):
+    #@wraps(Dataset.get, assigned=(), updated=())
+    # missing in dataset!
+    def get(self, sources, indexes):
```
fvisin: This seems like a but in the Dataset interface: _get_ is not part of the interface and I couldn't find it in iterator.py either.

bartvm: See #854 and #1364.

pylearn2/datasets/hdf5.py
```diff
@@ -2,338 +2,472 @@
...
-        return rval
+                sdata = self.data[s]
+            except ValueError as e:
+                reraise_as(ValueError(
+                    'The requested source %s is not part of the dataset' %
+                    sources[s], *e.args))
+            if isinstance(indexes, slice) or len(indexes) == 1:
+                rval.append(sdata[indexes])
+            else:
+                warnings.warn('Accessing non sequential elements of an '
```
bartvm: Does this not flood the console with warnings for each batch?

pylearn2/datasets/hdf5.py
```diff
@@ -2,338 +2,472 @@
...
+        if aliases is None:
+            aliases = ()
+            # set aliases to canonical name or to source name
+            for s in sources:
+                if s in ['X', 'topo_view']:
+                    aliases += ('features',)
+                elif s == 'y':
+                    aliases += ('targets',)
+                else:
+                    aliases += (s,)
```
lamblin: I think that there should be a flag that toggles that behaviour.
It would be on for the moment, but when we get rid of the hardcoded "targets" and "features" in the costs, we could turn it off. The documentation would say which automatic aliases are added.

pylearn2/datasets/hdf5.py
lamblin: My understanding was it had been deprecated a long time ago, and that `data_specs` should not be `None` in general.
I think DenseDesignMatrix allows it, to keep the historic behaviour, but there is no reason for other datasets to support it.

pylearn2/datasets/hdf5.py
lamblin: Yes, I think that this is what we should do. Where do you see that `'targets'` is passed but not used?

pylearn2/datasets/hdf5.py
```diff
@@ -2,338 +2,472 @@
...
+                                     return_tuple=return_tuple,
+                                     convert=convert)
+
+    def _get_canonical_sources(self):
+        """
+        Returns a list. The first element will be the source of `X` if present,
+        or of `topo_view` otherwise. The second element will be the source of
+        `y` if present. This is needed for compatibility with legacy code but
+        you should not rely on it.
+        """
```
lamblin: Oh, OK. I really don't think this is necessary to have this.
What is the problem if we don't, and force the user to always specify `data_specs` in `iterator()`?

pylearn2/datasets/hdf5.py
lamblin: Returning a list of spaces can make sense, especially if this is what the documentation tells.

pylearn2/datasets/hdf5.py
```diff
@@ -2,338 +2,472 @@
...
+            pass
+        return data
 
-class HDF5DatasetIterator(FiniteDatasetIterator):
+    #@wraps(Dataset.get_data_specs, assigned=(), updated=())
+    # not in dataset, but required by iteration.755
+    def get_data_specs(self, source_or_alias=None):
+        """
+            Returns a tuple `(space, source)` related to the provided
+            source_or_alias key, if any, or to the canonical source otherwise.
```
lamblin: I would say "all the sources" rather than "the canonical source".

pylearn2/datasets/hdf5.py
lamblin: Also, you could use `self.get_num_examples()` instead of `data[0].shape[0]`.

pylearn2/datasets/hdf5.py
lamblin: By default, warnings are only displayed once.

pylearn2/datasets/hdf5.py
fvisin: This line [DenseDesignMatrix](https://github.com/fvisin/pylearn2/blob/master/pylearn2/datasets/dense_design_matrix.py#L315) creates an iterator passing as argument of mode the shape of the `X` field. My code will pass the shape of the first element of data instead, so it will work if you call the iterator only on `y`. 

What seems strange to me is that when the iterator is called, `data_specs[1] == ('features', 'targets')` (provided that you have labels in the dataset). Shouldn't we call iterator with data_specs referring to a single source? Or, if we want to iterate on multiple sources at the same time, shouldn't we set mode according to both shapes? (I admit I didn't dig into mode, so I don't know what the shape is used for)

pylearn2/datasets/hdf5.py
fvisin: Ok, if you confirm this is the intended behavior I will remove this code. I will also delete the default value for the `data_specs` argument both here and in the Dataset interface, is it ok?

pylearn2/datasets/hdf5.py
fvisin: Yes you are right. This way of accessing the data should really be avoided, I intended it more for single values access (for instance to check something) than for batch iteration, but still I agree that it is better to avoid the warnings flood. 

What if I add a flag to stop warning after the first 10 iterations?

pylearn2/datasets/hdf5.py
fvisin: When we get rid of the hardcoded sources names we can simply get rid of these lines too. The rest of the code can work with aliases = None (I just checked and committed some minor modifications to allow it)

pylearn2/datasets/hdf5.py
fvisin: If your dataset has data that are not sources, loading all the keys of the dataset would break the compatibility with legacy code. This apply, for instance, to the case in which the default settings for the dataset are stored in the dataset itself.

pylearn2/datasets/hdf5.py
fvisin: There is a reference in the docstring but there is not a `def get` method in Dataset. Shouldn't it be added?

pylearn2/datasets/hdf5.py
fvisin: If the default intended behavior is that, no problem at all! I was supporting what seemed to be the current interface.

fvisin:
I don't understand the errors in Travis. Can you give me some insights?

pylearn2/datasets/dataset.py
```diff
@@ -94,14 +171,19 @@ def adjust_for_viewer(self, X):
...
         """ Returns true if the dataset includes targets """
-
         raise NotImplementedError()
 
     def get_topo_batch_axis(self):
         """
         Returns the index of the axis that corresponds to different examples
         in a batch when using topological_view.
+        WARNING: This method is deprecated and will be unsupported after 27
+                 July 27, 2015. Some classes, e.g. DenseDesignMatrix, might
```
bartvm: Travis doesn't like the fact that you suddenly use a different indentation level here.

pylearn2/datasets/hdf5_deprecated.py
```diff
@@ -0,0 +1,362 @@
...
+        Order of axes in topological view.
+    """
+
+    def topo_view_to_design_mat(self, V):
+        """
+        Generate a design matrix from the topological view.
+
+        This override of DefaultViewConverter.topo_view_to_design_mat does
+        not attempt to transpose the topological view, since transposition
+        is not supported by HDF5 datasets.
```
bartvm: You need a parameters section here as in
```
Parameters
----------
V : ndarray
    Whatever V is supposed to be.
```

pylearn2/datasets/hdf5.py
```diff
@@ -2,338 +2,485 @@
...
-                              "X_labels). Use X_labels at your own risk.")
-
-        if self.y_labels is not None:
-            assert self.y is not None
-            assert self.y.ndim <= 2
-            if self.load_all:
-                assert np.all(self.y < self.y_labels)
+        Class constructor
+        """
+        assert isinstance(filename, basestring)
```
bartvm: `basestring` doesn't exist in Python 3. Use `theano.compat.six.string_types` instead.

bartvm:
I commented inline on the errors.

fvisin:
When everything is ok to merge, let me know and I will rebase -i and squash the last few commits into the initial ones so that history doesn't get messy.

pylearn2/datasets/dataset.py
fvisin: Done.

pylearn2/datasets/hdf5.py
fvisin: Done.

pylearn2/datasets/hdf5.py
lamblin: I still think the aliases behaviour can be useful, especially if you define an experiment with some names in your model and costs, and then want to use a different hdf5 file where datasets have been named differently.

pylearn2/datasets/hdf5.py
lamblin: OK, then I would say "all declared sources", or something like that.

pylearn2/datasets/hdf5.py
lamblin: That's right, your code currently works. I'm just pointing out that you already have that code factored out in `self.get_num_examples()`, which is also part of the `Dataset` interface, so it would be more legible to use it. Also, if `get_num_examples` gets upgraded in the future (let's say, to perform so checking and cache the value), then you would get those benefits here as well.
It's jut a minor point, though.

When the iterator is called, we iterate jointly on as many sources as needed. This is essential if we want to keep the targets associated with the right inputs, for instance.
One assumption that is made by FiniteDatasetIterator, that could be more explicit, is that all sources have the same number of elements, and that elements at a given index in different sources always belong to the same example.

pylearn2/datasets/hdf5.py
lamblin: I think you can remove this code.
However, I don't think you can remove the default value for `data_specs`, as it comes after `mode`, which can have a default value. It is OK, however, to have a default value that is not valid in some classes.

pylearn2/datasets/hdf5.py
fvisin: Yes, I was missing that assumption but now I got it (I will remember it for the next time: never write comments on a PR on Friday night! :P)

For what concerns `self.get_num_examples()` I totally agree.

pylearn2/datasets/hdf5.py
fvisin: Ok, I will add an assert to prevent it to be None. I will also add a comment  both in `HDF5.py` and in `Dataset.py` specifying that we kept it for compatibility with legacy code but it should not be None.

pylearn2/datasets/hdf5.py
fvisin: If you remove this code you just remove the _default aliases_ part, the rest will still work as planned.

pylearn2/datasets/hdf5.py
fvisin: Ok

fvisin:
Ok, it is difficult to follow the inline discussions. I try to wrap up the issues that are still open.

1. We are not going to allow `data_specs = None` anymore. If so, I will delete the default value, switch the order of `data_specs` and `mode` in the arguments (`data_specs` is not part of the interface, so the order of the arguments of the iterator is up to us) and I will also get rid of the 3 `_get_canonical_something` functions, since I used them only for this. Is it ok?
2. The `__iter__` method is specific to the `DenseDesignMatrix`, so it should not be part neither of the interface (i.e. Dataset.py) nor of the `HDF5Dataset` class. Can you confirm it?
3. `get_data_specs` is called by [iteration](https://github.com/fvisin/pylearn2/blob/hdf5_remake/pylearn2/utils/iteration.py#L755). Should it be part of the interface (i.e. `dataset.py`)?
4. `get` should be part of the interface. ATM there is a reference in the docstring of `iteration.py`, but there is no reference at all in `dataset.py`. Should I add a `def get` method?
5. It is not clear to me what the input of `get` is supposed to be: what is the type of index when it is not a slice? The docstring says "a list or slice of indexes". Shouldn't it be a list of lists of indexes to cover the n-D case (e.g. [[0,0,0,0], [0,0,0,4], [0,2,0,1]])? What if the index is only one? Should we expect [[0,0,0,0]] or [0,0,0,0]?
6. How are we going to deal with the `aliases==None` scenario? Right now the code substitutes it with the _canonical sources_ to be compatible with legacy code. __But__ this is the new constructor: there is no legacy code! Who is going to use this constructor with legacy code will do it on purpose. This means that he can comply to the new interface (it is as simple as setting the `aliases` parameter calling the constructor).
7. With not too much work it will be possible to add the code to load the default settings (e.g. axes, cache_size, etc, ..). I don't have time to do it right now and it would be better to do it when the parameters will be more clear. Should I add a feature request for this?
8. A test in Travis is still failing: `pylearn2.utils.test_serial.test_load_train_file`. The problem is that it tries to load `test_model.yaml` that defines a dataset without parameters. I don't see how it could not fail before and, more importantly, if I should change something in the wrapper or the test should be modified. Insights?

lamblin:
On Mon, Feb 02, 2015, Francesco wrote:
> 1. We are not going to allow `data_specs = None` anymore. If so, I
> will delete the default value, switch the order of `data_specs` and
> `mode` in the arguments (`data_specs` is not part of the interface,
> so the order of the arguments of the iterator is up to us) and I will
> also get rid of the 3 `_get_canonical_something` functions, since I
> used them only for this. Is it ok?

data_specs should be in the interface, I will try to update the docstring
in `Dataset`. For the moment, I think we should keep the default of None.

Completely forbid data_specs=None could be done later.

> 2. The `__iter__` method is specific to the `DenseDesignMatrix`, so it
> should not be part neither of the interface (i.e. Dataset.py) nor of
> the `HDF5Dataset` class. Can you confirm it?

That is a good question. I did not even remember that this method existed.
It may be the reason why DenseDesignMatrix supported data_specs=None,
actually.

A way of checking if it is actually used would be to remove it and see
which tests fail. This could be another ticket.
If we decide to keep it (it could be useful, I guess), it could call
iterator with the default (native, internal) data_specs.

> 3. `get_data_specs` is called by [iteration](https://github.com/fvisin/pylearn2/blob/hdf5_remake/pylearn2/utils/iteration.py#L755).
> Should it be part of the interface (i.e. `dataset.py`)?

This is the case for FiniteDatasetIterator, not for all possible iterators.
If we add a `FiniteDataset` class, `get_data_specs` may be part of its
interface, but I don't think it should be in the base `Dataset` class.

> 4. `get` should be part of the interface. ATM there is a reference in
> the docstring of `iteration.py`, but there is no reference at all in
> `dataset.py`. Should I add a `def get` method?

Again, maybe for a `FiniteDataset` or `IndexableDataset`, but not
for all `Dataset`s.

> 5. It is not clear to me what the input of `get` is supposed to be:
> what is the type of index when it is not a slice? The docstring
> says "a list or slice of indexes". Shouldn't it be a list of lists
> of indexes to cover the n-D case (e.g. [[0,0,0,0], [0,0,0,4],
> [0,2,0,1]])? What if the index is only one? Should we expect
> [[0,0,0,0]] or [0,0,0,0]?

`get` is supposed to return a minibatch of examples. Its input is whatever
can identify those examples: a slice, or a list of example indices.
I think we can suppose all the examples are numbered by integer indices,
if some datasets index their examples in more than 1D, then they would
need a different iterator.

> 6. How are we going to deal with the `aliases==None` scenario? Right
> now the code substitutes it with the _canonical sources_ to be
> compatible with legacy code. __But__ this is the new constructor:
> there is no legacy code!
> This means that he can comply to
> the new interface (it is as simple as setting the `aliases` parameter
> calling the constructor).

Sure, we can do this before the constructor is called. But then,
there should be a warning about it, so when we drop support for the
legacy dataset, people are not left with broken scripts.

> 7. With not too much work it will be possible to add the code to load
> the default settings (e.g. axes, cache_size, etc, ..). I don't have
> time to do it right now and it would be better to do it when the
> parameters will be more clear. Should I add a feature request for
> this?

Sure, a follow-up ticket would be a good idea.

> 8. A test in Travis is still failing:
> `pylearn2.utils.test_serial.test_load_train_file`. The problem is that
> it tries to load `test_model.yaml` that defines a dataset without
> parameters. I don't see how it could not fail before and, more
> importantly, if I should change something in the wrapper or the test
> should be modified. Insights?

It is currently perfectly valid to instanciate a Dataset base class,
as long as you do not try to call its method.
If you remove the `__metaclass__` thing, it would probably fix the
test. For the moment, I think it is the right thing to do.
Let's discuss enforcing interfaces and implementations in two weeks.

-- 
Pascal

fvisin:
> data_specs should be in the interface, I will try to update the docstring in `Dataset`. For the moment, I think we should keep the default of None. Completely forbid data_specs=None could be done later.

It could be actually a good idea to be able to just get a "default iterator", i.e. an iterator on all the aliases (if any, or sources else) of the dataset. If you agree on this, we could just leave the default value `data_specs=None` with the meaning "refer to sources and spaces provided when the dataset object was created".

> That is a good question. I did not even remember that this method existed. It may be the reason why DenseDesignMatrix supported data_specs=None, actually. A way of checking if it is actually used would be to remove it and see which tests fail. This could be another ticket. If we decide to keep it (it could be useful, I guess), it could call iterator with the default (native, internal) data_specs.

I will remove it and try to see if any test fails. In case none does, I would get rid of this method (if we agree on the previous point, it would not add anything to the iterator method). 

> This is the case for FiniteDatasetIterator, not for all possible iterators. If we add a `FiniteDataset` class, `get_data_specs` may be part of its interface, but I don't think it should be in the base `Dataset` class.

Ok.

> Again, maybe for a `FiniteDataset` or `IndexableDataset`, but not for all `Dataset`s.

In this case, I would say that every dataset can be supposed to have a get method to get the data. It does not depend on how is the dataset and (source, index) seems pretty generic to fit every dataset.

>  `get` is supposed to return a minibatch of examples. Its input is whatever can identify those examples: a slice, or a list of example indices. I think we can suppose all the examples are numbered by integer indices, if some datasets index their examples in more than 1D, then they would need a different iterator.

I don't think that a different iterator is needed. The `get` method can easily check if an index is a tuple/list or an int and behave accordingly. I will write the code to deal with this.

> Sure, we can do this before the constructor is called. But then, there should be a warning about it, so when we drop support for the legacy dataset, people are not left with broken scripts.

I don't think I made myself clear. If you have a legacy dataset/code you will use the deprecated hdf5 handler by default (as per the `__new__` method) and it will warn you to switch to the new one. If you switch to the new one, you can easily specify the aliases (e.g. it is what I am doing right now to use the old imagenet dataset with the new hdf5 loader). I can specify more explicitly in the description of the parameter that if you want to use old datasets you should just set 'features' and 'targets' as aliases)

> Sure, a follow-up ticket would be a good idea.

Ok

> It is currently perfectly valid to instanciate a Dataset base class, as long as you do not try to call its method. If you remove the `__metaclass__` thing, it would probably fix the test. For the moment, I think it is the right thing to do. Let's discuss enforcing interfaces and implementations in two weeks.

Ok.

lamblin:
On Mon, Feb 02, 2015, Francesco wrote:
> It could be actually a good idea to be able to just get a "default
> iterator", i.e. an iterator on all the aliases (if any, or sources
> else) of the dataset. If you agree on this, we could just leave the
> default value `data_specs=None` with the meaning "refer to sources and
> spaces provided when the dataset object was created".

That makes sense.

> I will remove it and try to see if any test fails. In case none does,
> I would get rid of this method (if we agree on the previous point, it
> would not add anything to the iterator method).

Having an `__iter__` method actually adds something, it enables the
following syntax:
```python
for data in dataset:
    ...
```
rather than:
```python
it = dataset.iterator()
for data in it:
    ...
```
So I would not get rid of it.

> In this case, I would say that every dataset can be supposed to have a
> get method to get the data. It does not depend on how is the dataset
> and (source, index) seems pretty generic to fit every dataset.

That would not be true of datasets where the data is generated on the fly,
for instance. For other cases, `index` could be a hash, or a string.

> I don't think that a different iterator is needed. The `get` method
> can easily check if an index is a tuple/list or an int and behave
> accordingly. I will write the code to deal with this.

Right now, the convention is that the "mode" or iteration scheme in
FiniteDatasetIterator returns something that is fit to index an ndarray
or similar object, and is the only place using `get`.

If we want to use FiniteDatasetIterator everywhere, then we can
decide that all Datasets should have a `get` method that takes as
indices something suitable for it, but then the user has to make sure
it uses a mode that returns the right kind of indices for the dataset.
It would make more sense for me to have a different kind of Iterator
for these datasets, and if we do, we may as well not use `get`.

> I don't think I made myself clear. If you have a legacy dataset/code
> you will use the deprecated hdf5 handler by default (as per the
> `__new__` method) and it will warn you to switch to the new one. If
> you switch to the new one, you can easily specify the aliases (e.g. it
> is what I am doing right now to use the old imagenet dataset with the
> new hdf5 loader). I can specify more explicitly in the description of
> the parameter that if you want to use old datasets you should just set
> 'features' and 'targets' as aliases)

OK.
-- 
Pascal

fvisin:
> Having an `__iter__` method actually adds something, it enables the
> following syntax:
> ```python
> for data in dataset:
>     ...
> ```
> rather than:
> ```python
> it = dataset.iterator()
> for data in it:
>     ...
> ```
> So I would not get rid of it.

Agreed.

> That would not be true of datasets where the data is generated on the fly,
> for instance. For other cases, `index` could be a hash, or a string.

The interface seems still good to me, just to say that the `get` method is the one you should refer to when you want to get data from the dataset. The specific implementation will depend on the specific dataset. Index can be set to None by default: the on-the-fly dataset will not use it, the topo-view-like dataset will use a topo-view-like nd index, the design-matrix-like dataset will use an int. The common syntax would be that index is a list of _indices_, i.e., slice/int for dense design matrices, nd-slices/nd-int for topo-view, and so on ...

> Right now, the convention is that the "mode" or iteration scheme in
> FiniteDatasetIterator returns something that is fit to index an ndarray
> or similar object, and is the only place using `get`.

If _ndarray_ includes 1d arrays and 2d arrays, then you can have one code that deals with all the possible cases. The modification I was proposing would do exactly this, basically implementing the _list of indices_ idea of before, for the case of ndarrays (e.g. topo view and dense design matrices).

> If we want to use FiniteDatasetIterator everywhere, then we can
> decide that all Datasets should have a `get` method that takes as
> indices something suitable for it, but then the user has to make sure
> it uses a mode that returns the right kind of indices for the dataset.
> It would make more sense for me to have a different kind of Iterator
> for these datasets, and if we do, we may as well not use `get`.

If you prefer to keep them separated we need to think how to deal with topo-view-like datasets, cause ImageNet is in that format, but I don't think it is necessary.

I am probably loosing the big picture here, but I would say that the DenseDesign vs TopoView distinction is more semantic than topological: they both are nd arrays (the space in between is on purpose), what really matters to the end of training is the space that tells you the shape and the semantic of the data.

What really matters to be able to iterate on the data, instead, is just the shape of the dataset (that we already have). Once you pass a list of one or more indices of the same shape of the dataset, it is easy to check that they have the right shape and to manage the 3 different possible cases (slices, index, indices). The only situation in which this would not work, would be if you have a dataset whose data cannot be accessed specifying an index (whatever the format of the index), but I cannot imagine such a case.

pylearn2/datasets/dataset.py
```diff
@@ -2,21 +2,25 @@
...
+import warnings
+from pylearn2.utils.iteration import resolve_iterator_class
+
 
 class Dataset(object):
 
     """
     Abstract interface for Datasets.
     """
+    #__metaclass__ = abc.ABCMeta
```
lamblin: If assigning a metaclass does not work yet, we should remove that comment, and not import `abc`.

bartvm: You can assign a metaclass with the `@six.add_metaclass` decorator

pylearn2/datasets/dense_design_matrix.py
```diff
@@ -264,13 +264,14 @@ def _check_labels(self):
             assert self.y.ndim <= 2
             assert np.all(self.y < self.y_labels)
 
-    @functools.wraps(Dataset.iterator)
+    # Cannot wrap the iterator method cause it is abstract
+    # @functools.wraps(Dataset.iterator)
```
lamblin: This means that the docstring does not appear in the documentation anymore. It will probably trigger a test failure in the documentation tester.

pylearn2/datasets/hdf5.py
```diff
@@ -2,338 +2,467 @@
 Objects for datasets serialized in HDF5 format (.h5).
 """
 
-__author__ = "Steven Kearnes"
-__copyright__ = "Copyright 2014, Stanford University"
+__author__ = "Francesco Visin"
+__copyright__ = "Copyright 2015, Politecnico di Milano"
```
lamblin: I'm not sure about the copyright assignment, since you wrote this at UdeM.

pylearn2/datasets/hdf5.py
```diff
@@ -2,338 +2,467 @@
...
-        Parameters
-        ----------
-        WRITEME
+        data_specs : tuple, optional
+            A `(space, source)` tuple. See :ref:`data_specs` for a full
+            description. If set to None, the aliases (or sources) and spaces
+            provided when the dataset object has been created will be used.
+        return_tuple : bool, optional
+            Always return a tuple, even if there is exactly one source
+            of data being returned. Defaults to `False`.
```
lamblin: The content of that docstring seems partial. Is it supposed to get combined somehow with the docstring of `Dataset.iterator`?

pylearn2/datasets/hdf5.py
bartvm: Why do we add copyright anyway? Don't all the files fall under the same BSD-3 license, which lists as author "Université de Montréal"? Copyright would suggest that you have some sort of exclusive rights over the code you wrote, but by submitting it to Pylearn2 you waive all of those rights to UdeM as part of the license I think.

pylearn2/datasets/hdf5.py
```diff
@@ -2,338 +2,467 @@
...
+            The data
         """
-        next_index = self._subset_iterator.next()
+        return tuple([self.data[s] for s in self._get_sources()])
 
-        # convert to boolean selection
-        sel = np.zeros(self.num_examples, dtype=bool)
-        sel[next_index] = True
-        next_index = sel
+    # @wraps(Dataset.get, assigned=(), updated=())
```
lamblin: Remove commented-out code.

pylearn2/datasets/sparse_dataset.py
```diff
@@ -102,7 +102,8 @@ def get_batch_topo(self, batch_size):
     def get_num_examples(self):
         return self.X.shape[0]
 
-    @wraps(Dataset.iterator)
+    # Cannot wrap the iterator method cause it is abstract
+    # @wraps(Dataset.iterator)
```
lamblin: Idem.

pylearn2/datasets/dataset.py
lamblin: OK for now.

pylearn2/datasets/hdf5.py
lamblin: IANAL, but I don't think by submitting code to Pylearn2 we waive our copyright. The original copyright holder still has rights to use their own code with a different license, for instance, and can prosecute for copyright violations.
By working for the University or studying there, however, we do, so most of the code is copyright "Université de Montréal".
Canonical/Ubuntu actually was under fire recently for asking that contributors to their open source repository transfer their intellectual property to Canonical.

In that particular case, I'm not sure what the status of Francesco's intellectual property actually is, but I'm pretty sure UdeM and Politectnico will not fight over it :) so it does not matter.

pylearn2/datasets/dataset.py
fvisin: Adding the metaclass leads to a test failure. In my opinion the test is wrong (it instantiates an empty dataset), but apparently it is not. Anyway, I will remove this line.

pylearn2/datasets/dense_design_matrix.py
fvisin: Keeping the `functools.wraps` results in an error because it tries to call `Dataset.iterator`, which is an abstract method. 

You should either get rid of the abstract decorator, falling back in having a fake interface that is really just a class and as such can be instantiated, its method can be called, etc, .., or find a better way to inherit docstrings. 

Isn't there a way to specify it to the documentation generation tool? That would be the best option in my opinion. Another option is to add \__doc\__ += Dataset.iterator.\__doc\__ or use a more sophisticated way like writing your own decorator that inherits the docstring without actually calling the inherited method.

pylearn2/datasets/hdf5.py
fvisin: Yes, I specified just the parameter that are not present in the inherited method.

pylearn2/datasets/hdf5.py
fvisin: Removed the copyright

pylearn2/datasets/hdf5.py
fvisin: Ok

pylearn2/datasets/sparse_dataset.py
fvisin: Ok.

pylearn2/datasets/dense_design_matrix.py
lamblin: Updating `@pylearn2.utils.wraps` may be the way to go in the long run, or find a way to make NumPyDoc work with an abstract method.
But for the moment, let's do one thing at a time. First, let's merge the new version of HDF5Dataset, with inherited docstrings that work and show up in the online doc. Then we can impose an interface on Dataset and make sure it works nicely with docstrings, tests, and online doc.

fvisin: Ok. 
These could be good starting points to address the issue:
http://code.activestate.com/recipes/576862-docstring-inheritance-decorator/
http://code.activestate.com/recipes/578587-inherit-method-docstrings-without-breaking-decorat/

lamblin: Thanks, I opened a new issue at #1410

fvisin:
Ok, I rebased on master and cleaned history. I should also have implemented all the requested changes, let me know if something is still missing.

pylearn2/datasets/hdf5.py
lamblin: Did you generate the doc locally to make sure it works as intended?

lamblin: Also, the documentation for `Dataset.iterator` has already been updated to include `data_specs` and `return_tuple`.

fvisin: Ok, just pushed a new version. I modified the comment leaving only the part that specifies how I deal with  data_specs==None and I checked that it works as intended.

pylearn2/datasets/dataset.py
```diff
@@ -110,6 +108,79 @@ def iterator(self, mode=None, batch_size=None, num_batches=None,
...
+            details on the accepted seed formats. If unspecified,
+            defaults to using the dataset's own internal random
+            number generator, which persists across iterations
+            through the dataset and may potentially be shared by
+            multiple iterator objects simultaneously (see "Notes"
+            below).
+        data_specs : tuple
+            A `(space, source)` tuple that specifies the target space and
+            source that the iterator should provide. See :ref:`data_specs` i
+            for a full description.
```
lamblin: Could you re-use the docstring of `Dataset.iterator` for the `data_specs` argument? It explains that it needs to be flat, and what that means.

pylearn2/datasets/dataset.py
```diff
@@ -110,6 +108,79 @@ def iterator(self, mode=None, batch_size=None, num_batches=None,
...
+
+        Returns
+        -------
+        iter_obj : object
+            An iterator object implementing the standard Python
+            iterator protocol (i.e. it has an `__iter__` method that
+            return the object itself, and a `next()` method that
+            returns results until it raises `StopIteration`).
```
lamblin: Actually, that is not what this method returns.

fvisin: Wow...bad merge, my fault. Great catch!

pylearn2/datasets/dataset.py
fvisin: Ok, done.

pylearn2/datasets/dataset.py
```diff
@@ -110,6 +111,88 @@ def iterator(self, mode=None, batch_size=None, num_batches=None,
         # can be handled here.
         raise NotImplementedError()
 
+    def _init_iterator(self, mode=None, batch_size=None, num_batches=None,
+                       rng=None, data_specs=None):
+        """
+        Return an iterator for this dataset with the specified
+        behaviour. Unspecified values are filled-in by the default.
```
lamblin: It is not returning an iterator, it is preparing the arguments to pass to the `iterator` method.

pylearn2/datasets/dataset.py
```diff
@@ -110,6 +111,88 @@ def iterator(self, mode=None, batch_size=None, num_batches=None,
...
+            * Each corresponding (sub-space, source name) pair must be
+              unique, but the same source name may be mapped to many
+              sub-spaces (for example if one part of the model is fully
+              connected and expects a `VectorSpace`, while another part is
+              convolutional and expects a `Conv2DSpace`).
+
+            If `data_specs` is not provided, the behaviour (which
+            sources will be present, in which order and space, or
+            whether an Exception will be raised) is not defined and may
+            depend on the implementation of each `Dataset`.
```
lamblin: It still needs a `Returns` section, I think.

lamblin:
Otherwise, I think it is OK.

pylearn2/datasets/dataset.py
fvisin: Fixed (both) and rebased.

lamblin:
Thanks a lot!

fvisin:
Great, thank you!
Il 12/Mar/2015 14:02 "Pascal Lamblin" <notifications@github.com> ha scritto:

PR #1377, opened 2015-01-30
Update plot_monitor.py

jeremyfix:
I propose to add an option to set the yrange used for plotting the data ,to be used as :
      plot_monitor.py  model.pkl    --yrange min:max
e.g. plot_monitor.py  model.pkl    --yrange 0:1


lamblin:
Makes sense. Thanks!

PR #1378, opened 2015-02-02
plot_monitor.py

matrogers:
Sorry this took so long.  I had a lot of trouble getting the vertical alignment to work properly.  Although line 302 seems to align everything properly, I had to play around quite a bit to come up with 1-6*pad_height/total_height.  I'm fairly certain that this is not the best solution.

The pictures that it outputs are reasonably nice, but the dynamic display is still cutoff, and I'm not sure how to fix this without calling a package to add scrollbars.

test


lamblin:
Thanks!
For reference, this is for issue #962.

dwf:
@matrogers Can you fix the build failure? @lamblin will review.

PR #1384, opened 2015-02-05
Remove method todo write me. They are on Op methods documented in Theano...

nouiz:
.... We should not ducument all of their implementation.

### DIFF
--- a/pylearn2/sandbox/cuda_convnet/filter_acts.py
+++ b/pylearn2/sandbox/cuda_convnet/filter_acts.py
@@ -88,11 +88,6 @@ class FilterActs(BaseActs):
     # in FilterActs, that considers these parameters.
 
     def make_node(self, images, filters):
-        """
-        .. todo::
-
-            WRITEME
-        """
         if not isinstance(images.type, CudaNdarrayType):
             raise TypeError("FilterActs: expected images.type to be CudaNdarrayType, "
                     "got "+str(images.type))
@@ -133,11 +128,6 @@ class FilterActs(BaseActs):
         return flops
 
     def c_code(self, node, name, inputs, outputs, sub):
-        """
-        .. todo::
-
-            WRITEME
-        """
         images, filters = inputs
         targets, = outputs
         fail = sub['fail']
@@ -323,19 +313,9 @@ class FilterActs(BaseActs):
         return rval
 
     def c_code_cache_version(self):
-        """
-        .. todo::
-
-            WRITEME
-        """
         return (10,)
 
     def R_op(self, inputs, evals):
-        """
-        .. todo::
-
-            WRITEME
-        """
         images, filters = inputs
         images_ev, filters_ev = evals
         if 'Cuda' not in str(type(images)):
@@ -355,11 +335,6 @@ class FilterActs(BaseActs):
         return [sol]
 
     def grad(self, inputs, dout):
-        """
-        .. todo::
-
-            WRITEME
-        """
         images, filters = inputs
 
         if 'Cuda' not in str(type(images)):
--- a/pylearn2/sandbox/cuda_convnet/img_acts.py
+++ b/pylearn2/sandbox/cuda_convnet/img_acts.py
@@ -147,19 +147,9 @@ class ImageActs(BaseActs):
         return flops
 
     def connection_pattern(self, node):
-        """
-        .. todo::
-
-            WRITEME
-        """
         return [[1], [1], [0]]
 
     def grad(self, inputs, g_outputs):
-        """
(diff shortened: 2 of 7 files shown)


bartvm:
This will crash Travis though when these files are removed from the formatting whitelist (because the format checkers makes sure that every method, class, module, etc. has a docstring).

nouiz:
So this seem that we need to update those scripts to not check those methods. I propose that this is done when we remove those file from the list as this polluse the doc on the web currently:

http://deeplearning.net/software/pylearn2/library/alex.html?highlight=convnet

bartvm:
I agree, but I think that the blanket requirement of forcing each object to have a docstring might be counter-productive in general. It seems to have mostly lead to large amount of "WRITEME" statements, so maybe we should consider disabling it altogether instead of making an exception for these methods.

nouiz:
I think added the WRITEME was done fast due to the volume of change needed.
I think it is good that by default we expect them to be documented.
Otherwise, it is too easy to do nothing and have no warning/errors that the
doc is missing.

So I think a list of methods that do not need to be documented is good.

I would filter on more particular then the method node. I would make this
case, if the class is an instannce of theano.Op and for those method name,
disable the error.

bartvm:
I agree that warnings are a good idea, but I've always thought that making Travis fail seems a bit extreme. You could get a warning for each PR that introduces methods without docstrings using e.g. scrutinizer or landscape.io. Else we'd end up with two lists: A list of files that *should* have docstrings but haven't been fixed yet, and a list of files that don't need docstrings... Seems like a lot of overhead.

nouiz:
What I hade in mind to not request any other per file white list. What I
suggest is that the script that check the docstring itself have exception
of Theano Op subclass. That way it will apply to all file it check.

Fred

lamblin:
Ideally, those methods would use `@wraps` to refer to the docstring of the base class (`Op` or `COp` or `GPUOp` I guess). However, that documentation in Theano is not in a compatible format, if it is even inherited from `Op`.
- long (cleaner?) solution: make a Pylearn2 Op subclass, with all these docstrings implemented, and the documentation pointing at the online documentation for theano Ops.
- maybe more general solution: have a special token (such as `#noqa` for pep8) that means this particular docstring is not needed, and not to treat it as an error.

I don't really like the idea of hard-coding a particular class in the script itself, or having a separate whitelist of complete files that do not need docstrings.

nouiz:
Otherwise, move them to Theano?

Anyway, I think we should merge this PR now as it put pylearn2 in a better
state then without this PR.

lamblin:
Yes, moving them to Theano would work.
If we don't, however, I'm not sure this change is better, as removing the WRITEME makes it harder to find the methods that need documentation (even if that documentation is just a link to Theano's doc).

nouiz:
I think that those method should not end up in the user documentation as is
the case currently. Even if just for a link to theano doc.

Anyway, I let you decide if you prefer to keep it like now or merge this PR.

lamblin:
I don't see any way currently of not making some methods show up.
However, methods like `make_node` (at least their parameters) should be documented, as well as the parameters for the class itself. For the other ones, I think I'll hide the "todo" box from the website, but keep the "writeme" string for grepping purposes.

goodfeli:
@lamblin is right that using the wraps decorator is the right way to handle this.

I'm not totally sure why Theano isn't documented in a compatible format---this page is meant to be applicable to all code in the lab:
http://deeplearning.net/software/pylearn/v2_planning/API_coding_style.html
and it says to use numpydoc.

I didn't make up that page, I was told Pylearn was meant to follow it because it's a lab project. So that was what I enforced when I supervised Alassane on making the format checker.

Personally I think Theano should use the format checker too, though the whitelist is going to have to be gigantic initially. I think enforcement of formatting rules in Theano is very inconsistent and it has become a sort of political tool akin to a filibuster, i.e. if someone doesn't like something but doesn't have a valid argument against it they just invent new formatting rules.

nouiz:
That was a proposal of where we want to go for new software.

When we tried to use it, we hit bug in numpydoc, so it wasn't done. When
working on pylearn2 doc enforcement, they the bug show up again. But
someone from the lab fixed them.

So now, it would be possible to make Theano support that syntax. It is just
a question of time. No timeline for this.

As you do not seem to clean up the current web site with a quick fix, I'll
close this PR and let this to later. I won't got to fix all Theano for
pylearn2 doc.

There isn't any filibuster stuff in Theano. Just no enforcement cause the
current state + the fact that nobody remember the syntax that we wanted to
use but generated bad doc (so we didn't used it).

PR #1386, opened 2015-02-06
Update doc for data_specs in Dataset.iterator()

lamblin:


lamblin:
@goodfeli Could you have a look? Could you merge if it is clear enough, or let me know if it is not?

doc/conf.py
```diff
@@ -50,7 +50,7 @@
 
 # General substitutions.
 project = 'Pylearn2'
-copyright = '2011, LISA lab'
+copyright = '2011--2015, LISA lab'
```
goodfeli: should just be one hyphen

pylearn2/datasets/dataset.py
```diff
@@ -18,7 +18,7 @@ def __iter__(self):
         return self.iterator()
```
goodfeli: This file is still on the format checker's whitelist, both for docstrings and for PEP8. Can you take it off the whitelist so the formatting will be checked automatically? I don't think I can guarantee PEP8 and NumpyDoc compliance by manually reviewing the file.

lamblin:
@goodfeli Updated.

pylearn2/datasets/dataset.py
```diff
@@ -48,6 +48,28 @@ def iterator(self, mode=None, batch_size=None, num_batches=None,
...
             multiple iterator objects simultaneously (see "Notes"
             below).
+        data_specs : (space, source) pair, optional
+            `space` is an instance of `Space` (possibly a
+            `CompositeSpace`), and `source` is a string (or tuple of
+            strings if `space` is a `CompositeSpace`). `source` defines
+            which data should be returned by the iterator (for instance
+            'features', 'targets'), and `space` specifies in which
+            format it will be returned. These specifications should be
+            flat: if `space` is a `CompositeSpace`, the sub-spaces should
```
goodfeli: In some contexts, I've encountered a requirement that all Spaces be CompositeSpace at the root. Is that applicable here?

pylearn2/datasets/dataset.py
lamblin: I don't think we should be required to take files off the whitelist every time we update the documentation, but sure, I'll try to find time for that.

pylearn2/datasets/dataset.py
```diff
@@ -48,6 +48,28 @@ def iterator(self, mode=None, batch_size=None, num_batches=None,
...
+            not be `CompositeSpace`s, and `source` should not be a nested
+            tuple, and the same sub-source should not be mentioned more
+            than once with the same sub-space. If `data_specs` is not
+            provided, the behaviour (which sources will be present, in
+            which order and space, or whether an Exception will be raised)
+            is not defined and may depend on the implementation of each
+            `Dataset`.
```
goodfeli: why is it OK to not provide data_specs? Shouldn't we enforce this? are there any Dataset implementations that actually accept None?

pylearn2/datasets/dataset.py
```diff
@@ -48,6 +48,28 @@ def iterator(self, mode=None, batch_size=None, num_batches=None,
...
+            `space` is an instance of `Space` (possibly a
+            `CompositeSpace`), and `source` is a string (or tuple of
+            strings if `space` is a `CompositeSpace`). `source` defines
+            which data should be returned by the iterator (for instance
+            'features', 'targets'), and `space` specifies in which
+            format it will be returned. These specifications should be
+            flat: if `space` is a `CompositeSpace`, the sub-spaces should
+            not be `CompositeSpace`s, and `source` should not be a nested
+            tuple, and the same sub-source should not be mentioned more
+            than once with the same sub-space. If `data_specs` is not
```
goodfeli: "the same sub-source should not be mentioned more than once with the same sub-space"
By "with" do you mean "within"?
Isn't the global Space technically also a sub-space, and doesn't that imply that no sub-source  can be repeated?
Since the specification must be flat, doesn't that mean that `source` should be a list of unique strings?
It seems much simpler to say that source should be a list of unique strings than to have all this stuff about sub-spaces and sub-sources.

pylearn2/datasets/dataset.py
lamblin: `space` does not have to be a `CompositeSpace` here, however it is valid to request a `CompositeSpace` with only one component.

pylearn2/datasets/dataset.py
lamblin: `DenseDesignMatrix` supports `None`, and falls back to the `_iter_data_specs` private member for that.
As far as I recall, it was to avoid breaking existing tests.
The main benefit I can see is that the `iterator` method can be called with all default arguments, which makes it possible to use the dataset as an iterator itself, as implemented in `Dataset.__iter__()`.
I am not sure if that feature is worth having, but I think it is currently tested.

pylearn2/datasets/dataset.py
goodfeli: I hold myself to the same standard I am advocating here  e.g. I de-whitelisted utils/image.py after dwf raised concerns about PEP8 in one of my recent PRs ( https://github.com/lisa-lab/pylearn2/pull/1379 ) and the main point of that PR wasn't documentation.

pylearn2/datasets/dataset.py
goodfeli: What are the contexts where the space does have to be CompositeSpace? I remember encountering error messages relating to that while implementing models that only require a single variable. For example, sgd.py and monitor.py seem to require that the root be CompositeSpace. Are there other cases like that, other than things that are obviously intrinsically composite like CompositeLayer?

If Dataset.iterator accepts non-CompositeSpace roots, why do the other contexts have that restriction?

pylearn2/datasets/dataset.py
goodfeli: This observation is worth adding to the Notes section.

pylearn2/datasets/dataset.py
goodfeli: As a reminder, this is the e-mail thread with the original questions that need to be cleared up: https://groups.google.com/forum/#!topic/pylearn-dev/2PMNT1GEvCE

pylearn2/datasets/dataset.py
lamblin: It is not that all sources have to be unique, it is that all elementary (source, space) must be unique.
For instance, if two different costs want the same source (`'targets'`) in different formats (one-hot vs. index), in addition to the features, then the source strings are not unique: `'targets'` will be specified twice, once in conjunction with a `VectorSpace`, and once with an `IndexSpace`.
Much of the complexity of the `data_specs` come from the requirement that the `source` only indicates _what_ data, and not the format (so that swapping models for others with different input spaces can be done without changing the name of the sources they are requiring), and that different parts in the same experiment can request the same data in different formats at the same time.
It would be good if we find a way to fulfill these requirements in a simpler way.

goodfeli: How about something like

  data_specs : (space, source) pair
        `space` must be an instance of Space and `source` must be a string or tuple of string names such as 'features' or 'targets'. The source names specify where the data will come from and the Space specifies its format.
        When source is a tuple, there are some additional requirements:
          * `space` must be a CompositeSpace, with one sub-space corresponding to each source name. i.e., the specification must be flat.
          * None of the components of `space` may be a CompositeSpace
          * Each corresponding (sub-space, source name) pair must be unique, but the same source name may be mapped to many sub-spaces (for example if one part of the model is fully connected and expects a VectorSpace, while another part is convolutional and expects a Conv2DSpace)

pylearn2/datasets/dataset.py
lamblin: I am not aware of any contexts where a provided `Space` _must_ be a `CompositeSpace`. There are however contexts where all spaces created or _returned_ by a function will be instances of `CompositeSpace`, even if that is not required by the caller of that function.
I don't necessarily see a problem in being stricter in the value returned at these places, but more tolerant in the values used by `Dataset.iterator`, especially since the `iterator` method can sometimes be called from a user's script.

pylearn2/datasets/dataset.py
lamblin: That makes sense. I'll try to integrate that tomorrow.

pylearn2/datasets/dataset.py
lamblin: Done.
I had to update the signature of `get_batch_topo` to get consistant with the implementation in `DenseDesignMatrix`, and took most of the documentation from `get_batch_design`.
Ultimately, I think these methods should be deprecated and disappear, maybe being replaced by one that requests a list of sources and spaces.

pylearn2/datasets/dataset.py
lamblin: Done.

pylearn2/datasets/dataset.py
lamblin: Integrated.

lamblin:
I think all your comments should be addressed, unless something is still not clear about the use of `CompositeSpace`, in which case, please let me know.

pylearn2/datasets/dataset.py
goodfeli: get_batch_topo and get_batch_design are useful for quick and dirty experimentation.

While it is important for Pylearn2 to be flexible, it's also important for it to be easy to use. If I want to write a script that just loads a design matrix and runs it through an autoencoder's reconstruction function or something like that, I don't want to spend two hours reading the documentation about spaces and sources.

Rather than getting rid of get_batch_design and get_batch_topo, I'd suggest making them methods only of DenseDesignMatrix.

lamblin: That is what I meant, they should disappear from the interface of `Dataset`, but can still be part of `DenseDesignMatrix`.

PR #1399, opened 2015-02-11
[WIP] Add cuda_convnet local-connected wrapper. 

nouiz:
TODO:

- [ ] need to finish the tests. They compare again the old results. We need to add the expected value in the new case.
- [ ] code clean up


lamblin:
Is it intended to replace the implementation in `pylearn2/packaged_dependencies/theano-linear/unshared_conv/`?

PR #1406, opened 2015-02-14
Support negative sampling

minhlab:
The main idea is one layer (with one set of weights and bias) produces two sets of complementary probabilities. One of them will be matched with positive examples, the other with negative random examples. I assume that one can generate those negative examples using a home-grown dataset or Pylearn2's transformer dataset. The default cost is negative log likelihood over both sets of probabilities.

### DIFF
--- a/pylearn2/sandbox/nlp/models/mlp.py
+++ b/pylearn2/sandbox/nlp/models/mlp.py
@@ -1,18 +1,24 @@
 """
 Sandbox multilayer perceptron layers for natural language processing (NLP)
 """
-import theano.tensor as T
-from theano import config
-
+from pylearn2.compat import OrderedDict
 from pylearn2.models import mlp
 from pylearn2.models.mlp import Layer
+from pylearn2.sandbox.nlp.linear.matrixmul import MatrixMul
+from pylearn2.space import CompositeSpace
+from pylearn2.space import Space
 from pylearn2.space import IndexSpace
 from pylearn2.space import VectorSpace
-from pylearn2.space import CompositeSpace
 from pylearn2.utils import sharedX
 from pylearn2.utils import wraps
-from pylearn2.sandbox.nlp.linear.matrixmul import MatrixMul
-from pylearn2.compat import OrderedDict
+
+from pylearn2.expr.nnet import arg_of_sigmoid
+from pylearn2.space import Conv2DSpace
+from pylearn2.utils import py_integer_types
+from theano import tensor as T
+
+import numpy as np
+import theano
 
 
 class Softmax(mlp.Softmax):
@@ -105,12 +111,12 @@ class ProjectionLayer(Layer):
         row_norms = T.sqrt(sq_W.sum(axis=1))
         col_norms = T.sqrt(sq_W.sum(axis=0))
 
-        return OrderedDict([('row_norms_min',  row_norms.min()),
+        return OrderedDict([('row_norms_min', row_norms.min()),
                             ('row_norms_mean', row_norms.mean()),
-                            ('row_norms_max',  row_norms.max()),
-                            ('col_norms_min',  col_norms.min()),
+                            ('row_norms_max', row_norms.max()),
+                            ('col_norms_min', col_norms.min()),
                             ('col_norms_mean', col_norms.mean()),
-                            ('col_norms_max',  col_norms.max()), ])
+                            ('col_norms_max', col_norms.max()), ])
 
     def _check_input_space_and_get_max_labels(self, space):
         if isinstance(space, IndexSpace):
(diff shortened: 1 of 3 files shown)


minhlab:
I'm not so happy with the name of the class. Please let me know if you have a better idea.

lamblin:
@bartvm Since this is in `sandbox/mlp`, could you have a look? Thanks.

pylearn2/sandbox/nlp/models/mlp.py
```diff
@@ -189,3 +198,303 @@ def get_l1_weight_decay(self, coeff):
         assert isinstance(coeff, float) or hasattr(coeff, 'dtype')
         W, = self.transformer.get_params()
         return coeff * abs(W).sum()
+
+
```
goodfeli: Is there a published reference for this anywhere?

pylearn2/sandbox/nlp/models/mlp.py
```diff
@@ -189,3 +198,303 @@ def get_l1_weight_decay(self, coeff):
...
+class ContrastiveProbabilityLayer(Layer):
+    """
+    Predicts the probability of a sample to come from a positive distribution
+    or a negative distribution. For each sample (i.e. a row in the input),
+    this layer produces *two* sets of probabilities. The first set has n
+    probabilities which says "if the sample is of class i then the probability
+    it comes from the positive distribution is p_i". The second set is the
+    complement of the first.
```
goodfeli: why output the redundant second set?

pylearn2/sandbox/nlp/models/mlp.py
```diff
@@ -189,3 +198,303 @@ def get_l1_weight_decay(self, coeff):
         assert isinstance(coeff, float) or hasattr(coeff, 'dtype')
         W, = self.transformer.get_params()
         return coeff * abs(W).sum()
+
+
+class ContrastiveProbabilityLayer(Layer):
+    """
```
goodfeli: How are these distributions defined?

pylearn2/sandbox/nlp/models/mlp.py
```diff
@@ -189,3 +198,303 @@ def get_l1_weight_decay(self, coeff):
...
+    Predicts the probability of a sample to come from a positive distribution
+    or a negative distribution. For each sample (i.e. a row in the input),
+    this layer produces *two* sets of probabilities. The first set has n
```
goodfeli: So you have a model p(x, c) and this is the function that outputs p(c | x) ?

pylearn2/sandbox/nlp/models/mlp.py
minhlab: By "this" you mean the L1 weight?

pylearn2/sandbox/nlp/models/mlp.py
minhlab: It's ugly, I know. The point was to let users provide their own set of positive and negative examples. The first set of output is compared to positive examples while the second set of output negative examples.

pylearn2/sandbox/nlp/models/mlp.py
minhlab: About negative sampling, maybe this helps: [http://arxiv.org/abs/1402.3722](http://arxiv.org/abs/1402.3722).

pylearn2/sandbox/nlp/models/mlp.py
minhlab: The distribution of positive and negative examples? Positive examples are taken from a dataset, negative ones are randomly generated. It can be as simple as a uniform distribution or corruption of positive examples. In Socher et al. (2013) they take a correct triple and replace the entity in either left hand side or right hand side. The replacement is constrained to be one that has appeared in the same relation and the same location.

Socher, R., Chen, D., Manning, C. D., & Ng, A. Y. (2013). Reasoning With Neural Tensor Networks for Knowledge Base Completion. Advances in Neural Information Processing Systems, 926–934.

pylearn2/sandbox/nlp/models/mlp.py
minhlab: No, if an example is (x, c) and it can come from either positive distribution D or negative distribution N then we are modeling P(D | x,c) and P(N | x,c).

pylearn2/sandbox/nlp/models/mlp.py
goodfeli: ctrl-F contrastive gets no hits.

goodfeli: and in this file, ctrl-F word2vec gets no hits. You really need to explain what this file is meant to do better.

pylearn2/sandbox/nlp/models/mlp.py
goodfeli: what does your docstring mean by "complement" specifically?

minhlab: something along this line: https://en.wikipedia.org/wiki/Complementary_event

pylearn2/sandbox/nlp/models/mlp.py
minhlab: It's not necessarily about word2vec, more and more papers are using negative sampling as well. For example Neural Tensor Network is trained with a similar scheme. Quick reference:

> In order to create a testing set for classification, we randomly switch entities from correct testing triplets resulting in a total of 2×#Test triplets with equal number of positive and negative examples.

Socher, R., Chen, D., Manning, C. D., & Ng, A. Y. (2013). Reasoning With Neural Tensor Networks for Knowledge Base Completion. Advances in Neural Information Processing Systems, 926–934.

minhlab: Yes, the word "contrastive" is obscure. Maybe I'd better call the class ComplementaryLayers.

minhlab: More about negative sampling:
http://scholar.google.nl/scholar?as_ylo=2015&q=%22negative+sampling%22&hl=en&as_sdt=0,5

minhlab:
Skip this one, I will make a better PR :-)

PR #1407, opened 2015-02-15
Use cudnn when possible instead of conv2d

fvisin:
This is a merge of an old work by @ballasn.

### DIFF
--- a/pylearn2/linear/cudnn2d.py
+++ b/pylearn2/linear/cudnn2d.py
@@ -1,0 +1,272 @@
+"""
+A module for convolutions with cudnn.
+"""
+
+__author__ = "Nicolas Ballas"
+__license__ = "3-clause BSD"
+__credits__ = "Nicolas Ballas and Francesco Visin"
+__maintainer__ = "Lisa Lab"
+
+import functools
+import numpy as np
+
+from theano.sandbox.cuda.dnn import GpuDnnConv, GpuDnnConvDesc
+from theano.sandbox.cuda.basic_ops import gpu_contiguous, gpu_alloc_empty
+
+from pylearn2.packaged_dependencies.theano_linear.conv2d \
+    import Conv2d as OrigConv2D
+
+from pylearn2.linear.linear_transform import LinearTransform as P2LT
+from pylearn2.utils import sharedX
+from pylearn2.utils.rng import make_np_rng
+
+
+default_seed = [2012, 11, 6, 9]
+default_sparse_seed = [2012, 11, 6]
+
+
+class Cudnn2D(OrigConv2D):
+    """
+    Wrapper on the Theano Cudnn op.
+
+    Parameters
+    ----------
+    filters : Theano shared variable
+        4D-tensor of shape (out channels, in channels, rows, cols)
+    batch_size : int
+        The size of the input batches
+    input_space : Space
+        The Space of the input data
+    output_axes : tuple, optional
+        The requested output axes. If not specified `bc01` will be used.
+    subsample : tuple or list, optional
+        Factor by which to subsample the output. Default (1, 1)
+    border_mode : string, optional
+        `valid` or `full`. See scipy.signal.convolve2d
+    filters_shape : tuple of length 2 or 3, optional
+        ([filter's number,] filter's height, filter's width)
+    message : string, optional
+        TODO
+    """
+
+    def __init__(self,
+                 filters,
+                 batch_size,
+                 input_space,
+                 output_axes=('b', 'c', 0, 1),
+                 subsample=(1, 1),
+                 border_mode='valid',
+                 filters_shape=None,
+                 message=''):
+
+        assert batch_size is None or batch_size > 0
+        self._input_space = input_space
+        self._output_axes = output_axes
+        self._subsample = tuple(subsample)
+        self._border_mode = border_mode
+
+        super(Cudnn2D, self).__init__(
+            filters=filters,
+            img_shape=(batch_size, input_space.num_channels,
+                       input_space.shape[0], input_space.shape[1]),
+            subsample=self._subsample,
+            border_mode=border_mode,
+            filters_shape=filters.get_value(borrow=True).shape,
+            message=message
+        )
+
+        # conv_op has to be changed
+        self._conv_op = GpuDnnConv()
+        self._desc = GpuDnnConvDesc(border_mode=border_mode,
+                                    subsample=self._subsample,
+                                    conv_mode='conv')
+
+    @functools.wraps(P2LT.get_params)
+    def get_params(self):
+        """ Return self._filters. """
+        return [self._filters]
+
+    @functools.wraps(P2LT.get_weights_topo)
+    def get_weights_topo(self, borrow):
+        """
+        Parameters
+        ----------
+        borrow : TODO
+            TODO
+        """
+        return np.transpose(self._filters.get_value(borrow=borrow),
+                            (0, 2, 3, 1))
+
+    def lmul(self, x):
+        """
+        .. todo::
+
+            WRITEME properly
+
+        dot(x, A)
+
+        This method overrides the original Conv2D lmul to make it work
+        with arbitrary axis orders
+
+        Parameters
+        ----------
+        x : TODO
+            TODO
+        """
+        # x must be formatted as batch index, channel, topo dim 0, topo dim 1
+        # for use with conv2d, so check what the current input space format is
+        assert x.ndim == 4
+        axes = self._input_space.axes
+        assert len(axes) == 4
+
+        op_axes = ('b', 'c', 0, 1)
+
+        if tuple(axes) != op_axes:
+            x = x.dimshuffle(*[axes.index(ax) for ax in op_axes])
+
+        # The calling format has to be changed
+        img = gpu_contiguous(x)
+        kerns = gpu_contiguous(self._filters)
+        shape = GpuDnnConv.get_out_shape(
+            img.shape, kerns.shape, self._border_mode, self._subsample)
+        rval = gpu_alloc_empty(*shape)
+        desc = self._desc(img.shape, kerns.shape)
+        rval = self._conv_op(img, kerns, rval, desc)
+
+        # Format the output based on the output space
+        axes = self._output_axes
+        assert len(axes) == 4
+
+        if tuple(self._output_axes) != op_axes:
+            rval = rval.dimshuffle(*[op_axes.index(ax) for ax in self._output_axes])
+
+        return rval
+
+    def set_batch_size(self, batch_size):
+        """
+        .. todo::
+
+            WRITEME
+
+        Parameters
+        ----------
+        batch_size : TODO
+            TODO
+        """
+        self._img_shape = tuple([batch_size] + list(self._img_shape[1:]))
+
+
+def make_random_conv2D(irange,
+                       input_space,
+                       output_space,
+                       kernel_shape,
+                       batch_size=None,
+                       subsample=(1, 1),
+                       border_mode = 'valid',
+                       message = "",
+                       rng = None):
+    """
+    .. todo::
+
+        WRITEME properly
+
+    Creates a CorrMM2D with random kernels
+
+    Parameters
+    ----------
+    irange : TODO
+    input_space : TODO
+    output_space : TODO
+    kernel_shape : 2D list or tuple
+    batch_size : int, optional
+    subsample : tuple, optional
+    border_mode : string, optional
+    message : string, optional
+    rng : optional
+    """
+
+    rng = make_np_rng(rng, default_seed, which_method='uniform')
+
+    W = sharedX(rng.uniform(-irange, irange, (output_space.num_channels,
+                                              input_space.num_channels,
+                                              kernel_shape[0],
+                                              kernel_shape[1])))
+
+    return Cudnn2D(
+        filters=W,
+        batch_size=batch_size,
+        input_space=input_space,
+        output_axes=output_space.axes,
+        subsample=tuple(subsample),
+        border_mode=border_mode,
+        filters_shape=W.get_value(borrow=True).shape,
+        message=message
+    )
+
+
+def make_sparse_random_conv2D(num_nonzero,
+                              input_space,
+                              output_space,
+                              kernel_shape,
+                              batch_size,
+                              subsample=(1, 1),
+                              border_mode='valid',
+                              message="",
+                              rng=None):
+
+    """
+    .. todo::
+
+        WRITEME properly
+
+    Creates a Cudnn2D with random kernels, where the randomly initialized
+    values are sparse
+
+    Parameters
+    ----------
+    num_nonzero : TODO
+    input_space : TODO
+    output_space : TODO
+    kernel_shape : TODO
+    batch_size : TODO
+    subsample : TODO, optional
+    border_mode : TODO, optional
+    message : TODO, optional
+    rng : TODO, optional
+
+    """
+
+    raise AssertionError(
+        "TODO: I think this is a bug--num_nonzero "
+        "determines the total number of nonzero elements in the "
+        "whole kernel stack, not the number of non-zero elements per "
+        "kernel. Investigate what it's meant to do."
+    )
+
+    rng = make_np_rng(rng, default_sparse_seed,
+                      which_method=['randn', 'randint'])
+
+    W = np.zeros((output_space.num_channels, input_space.num_channels,
+                  kernel_shape[0], kernel_shape[1]))
+
+    def random_coord():
+        return [rng.randint(dim) for dim in W.shape]
+
+    for i in range(num_nonzero):
+        o, ch, r, c = random_coord()
+        while W[o, ch, r, c] != 0:
+            o, ch, r, c = random_coord()
+        W[o, ch, r, c] = rng.randn()
+
+    W = sharedX(W)
+
+    return Cudnn2D(
+        filters=W,
+        batch_size=batch_size,
+        input_space=input_space,
+        output_axes=output_space.axes,
+        subsample=subsample,
+        border_mode=border_mode,
+        filters_shape=W.get_value(borrow=True).shape,
+        message=message
+    )
--- a/pylearn2/linear/cudnn2d.py
+++ b/pylearn2/linear/cudnn2d.py
@@ -138,7 +138,8 @@ class Cudnn2D(OrigConv2D):
         assert len(axes) == 4
 
         if tuple(self._output_axes) != op_axes:
-            rval = rval.dimshuffle(*[op_axes.index(ax) for ax in self._output_axes])
+            rval = rval.dimshuffle(*[op_axes.index(ax) for ax in
+                                     self._output_axes])
 
         return rval
 
--- a/pylearn2/linear/tests/test_cudnn.py
+++ b/pylearn2/linear/tests/test_cudnn.py
@@ -1,0 +1,168 @@
+"""
+Tests for the Cudnn code.
+"""
+
+__author__ = "Francesco Visin"
+__license__ = "3-clause BSD"
+__credits__ = "Francesco Visin"
+__maintainer__ = "Lisa Lab"
+
+import theano
+from theano import tensor
+from theano.sandbox.cuda.dnn import dnn_available
+
+from pylearn2.linear.conv2d import Conv2D
+from pylearn2.linear.cudnn2d import Cudnn2D, make_random_conv2D
+from pylearn2.space import Conv2DSpace
+from pylearn2.utils import sharedX
+from pylearn2.testing.skip import skip_if_no_gpu
+
+import unittest
+from nose.plugins.skip import SkipTest
+
+import numpy as np
+
+
+class TestCudnn(unittest.TestCase):
+    """
+    Tests for the Cudnn code.
+
+    Parameters
+    ----------
+    Refer to unittest.TestCase.
+
+    """
+    def setUp(self):
+        """
+        Set up a test image and filter to re-use.
+        """
+        skip_if_no_gpu()
+        if not dnn_available():
+            raise SkipTest('Skipping tests cause cudnn is not available')
+        self.orig_floatX = theano.config.floatX
+        theano.config.floatX = 'float32'
+        self.image = np.random.rand(1, 1, 3, 3).astype(theano.config.floatX)
+        self.image_tensor = tensor.tensor4()
+        self.input_space = Conv2DSpace((3, 3), 1, axes=('b', 'c', 0, 1))
+        self.filters_values = np.ones(
+            (1, 1, 2, 2), dtype=theano.config.floatX
+        )
+        self.filters = sharedX(self.filters_values, name='filters')
+        self.batch_size = 1
+
+        self.cudnn2d = Cudnn2D(self.filters, self.batch_size, self.input_space)
+
+    def tearDown(self):
+        """
+        After test clean up.
+        """
+        theano.config.floatX = self.orig_floatX
+
+    def test_value_errors(self):
+        """
+        Check correct errors are raised when bad input is given.
+        """
+        with self.assertRaises(AssertionError):
+            Cudnn2D(filters=self.filters, batch_size=-1,
+                    input_space=self.input_space)
+
+    def test_get_params(self):
+        """
+        Check whether the cudnn has stored the correct filters.
+        """
+        self.assertEqual(self.cudnn2d.get_params(), [self.filters])
+
+    def test_get_weights_topo(self):
+        """
+        Check whether the cudnn has stored the correct filters.
+        """
+        self.assertTrue(np.all(
+            self.cudnn2d.get_weights_topo(borrow=True) ==
+            np.transpose(self.filters.get_value(borrow=True), (0, 2, 3, 1))))
+
+    def test_lmul(self):
+        """
+        Use conv2D to check whether the convolution worked correctly.
+        """
+        conv2d = Conv2D(self.filters, self.batch_size, self.input_space,
+                        output_axes=('b', 'c', 0, 1),)
+        f_co = theano.function([self.image_tensor],
+                               conv2d.lmul(self.image_tensor))
+        f_cu = theano.function([self.image_tensor],
+                               self.cudnn2d.lmul(self.image_tensor))
+        self.assertTrue(np.allclose(f_co(self.image), f_cu(self.image)))
+
+    def test_set_batch_size(self):
+        """
+        Make sure that setting the batch size actually changes the property.
+        """
+        img_shape = self.cudnn2d._img_shape
+        self.cudnn2d.set_batch_size(self.batch_size + 10)
+        np.testing.assert_equal(self.cudnn2d._img_shape[0],
+                                self.batch_size + 10)
+        np.testing.assert_equal(self.cudnn2d._img_shape[1:], img_shape[1:])
+
+    def test_axes(self):
+        """
+        Test different output axes.
+
+        Use different output axes and see whether the output is what we
+        expect.
+        """
+        default_axes = ('b', 'c', 0, 1)
+        axes = (0, 'b', 1, 'c')
+        another_axes = (0, 1, 'c', 'b')
+        # 1, 3, 0, 2
+        map_to_default = tuple(axes.index(axis) for axis in default_axes)
+        # 2, 0, 3, 1
+        map_to_another_axes = tuple(default_axes.index(axis) for
+                                    axis in another_axes)
+        input_space = Conv2DSpace((3, 3), num_channels=1, axes=another_axes)
+        # Apply cudnn2d with `axes` as output_axes
+        cudnn2d = Cudnn2D(self.filters, 1, input_space, output_axes=axes)
+        f = theano.function([self.image_tensor],
+                            cudnn2d.lmul(self.image_tensor))
+        # Apply cudnn2d with default axes
+        f_def = theano.function([self.image_tensor],
+                                self.cudnn2d.lmul(self.image_tensor))
+
+        # Apply f on the `another_axes`-shaped image
+        output = f(np.transpose(self.image, map_to_another_axes))
+        # Apply f_def on self.image (b,c,0,1)
+        output_def = np.array(f_def(self.image))
+        # transpose output to def
+        output = np.transpose(output, map_to_default)
+
+        np.testing.assert_allclose(output_def, output)
+        np.testing.assert_equal(output_def.shape, output.shape)
+
+    def test_channels(self):
+        """
+        Go from 2 to 3 channels and see whether the shape is correct.
+        """
+        input_space = Conv2DSpace((3, 3), num_channels=3)
+        filters_values = np.ones(
+            (2, 3, 2, 2), dtype=theano.config.floatX
+        )
+        filters = sharedX(filters_values)
+        image = np.random.rand(1, 3, 3, 3).astype(theano.config.floatX)
+        cudnn2d = Cudnn2D(filters, 1, input_space)
+        f = theano.function([self.image_tensor],
+                            cudnn2d.lmul(self.image_tensor))
+        assert f(image).shape == (1, 2, 2, 2)
+
+    def test_make_random_conv2D(self):
+        """
+        Test a random convolution.
+
+        Create a random convolution and check whether the shape, axes and
+        input space are all what we expect.
+        """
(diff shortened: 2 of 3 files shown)


pylearn2/models/mlp.py
```diff
@@ -2987,15 +2995,27 @@ def initialize_transformer(self, rng):
...
+                    rng=rng)
+            else:
+                self.transformer = conv2d.make_random_conv2D(
+                    irange=self.irange,
+                    input_space=self.input_space,
+                    output_space=self.detector_space,
+                    kernel_shape=self.kernel_shape,
+                    subsample=self.kernel_stride,
+                    border_mode=self.border_mode,
+                    rng=rng)
```
lamblin: Here, we completely ignore `pad`, we should have a meaningful error message instead if `pad` is specified but `cudnn2d` is not available.

pylearn2/models/mlp.py
```diff
@@ -3059,6 +3079,14 @@ def set_input_space(self, space):
...
 
+        if self.pad != (0, 0):
+            output_shape = \
+                [int(np.ceil((i_sh + 2. * k_pad - k_sh) / float(k_st))) + 1
+                 for i_sh, k_sh, k_st, k_pad in izip(self.input_space.shape,
+                                                     self.kernel_shape,
+                                                     self.kernel_stride,
+                                                     self.pad)]
+
         if self.border_mode == 'valid':
```
lamblin: We should make sure 'border_mode' and 'pad' are not passed at the same time, I guess.
Also, the `output_shape` will be overwritten by default.

pylearn2/models/mlp.py
```diff
@@ -3186,6 +3214,25 @@ def get_weights_topo(self):
...
+        assert W.ndim == 4
+
+        sq_W = T.sqr(W)
+
+        row_norms = T.sqrt(sq_W.sum(axis=(1, 2, 3)))
+
+        return OrderedDict([('kernel_norms_min',  row_norms.min()),
+                            ('kernel_norms_mean', row_norms.mean()),
+                            ('kernel_norms_max',  row_norms.max()), ])
+
```
lamblin: That method had been removed, we should not put it back here.

pylearn2/models/mlp.py
```diff
@@ -3204,6 +3251,14 @@ def get_layer_monitoring_channels(self, state_below=None,
                            ('kernel_norms_max', row_norms.max()),
                            ])
 
+        if cudnn2d:
```
lamblin: Why is this needed?

pylearn2/linear/cudnn2d.py
```diff
@@ -0,0 +1,168 @@
...
+    """
+    Wrapper on the Theano Cudnn op.
+    """
+
+    def __init__(self, filters,
+                 batch_size,
+                 input_space,
+                 output_axes=('b', 'c', 0, 1),
+                 subsample=(1, 1),
+                 pad=(1,1),
```
lamblin: Why that for default padding, and not `(0, 0)`?

pylearn2/models/mlp.py
```diff
@@ -3059,6 +3079,14 @@ def set_input_space(self, space):
 
         rng = self.mlp.rng
 
+        if self.pad != (0, 0):
+            output_shape = \
+                [int(np.ceil((i_sh + 2. * k_pad - k_sh) / float(k_st))) + 1
```
lamblin: Actually, I'm not sure the `ceil` part is right, as it does not give the same result as the integer division below when pad is `(0, 0)`.

pylearn2/models/mlp.py
```diff
@@ -3246,9 +3301,19 @@ def fprop(self, state_below):
...
-                p = max_pool(bc01=d, pool_shape=self.pool_shape,
-                             pool_stride=self.pool_stride,
-                             image_shape=self.detector_space.shape)
+                # TODO Can they be merged?
+                if cudnn2d:
+                    # CPU implementation
+                    # TODO check if this can be ported to the GPU one
+                    # TODO check axes: should they be B01C or BC01 or will this
+                    # always work?
+                    p = downsample.max_pool_2d(d, self.pool_shape,
```
lamblin: You are not passing `self.pool_stride` here, which is why the dimensions are not right.

nouiz:
I just want to bring to your attention that the pooling operator is better handled (and have fix when stride > window size) in this PR: https://github.com/lisa-lab/pylearn2/pull/1361

Maybe you should remove the related change to this part? I just need to check if tests I ran passed to merge it.

pylearn2/models/mlp.py
fvisin: (I got rid of pad, following Fred's suggestion, so this comment does not apply anymore to this PR. I keep it for future reference.)

`border_mode` and `pad` can be passed together, as the latter augments the data but does not guarantee to do it such as the final size is a multiple of the kernel convolution (i.e., you might still need to either pad more or ignore part of the _padded image_)

pylearn2/models/mlp.py
fvisin: (I got rid of pad, following Fred's suggestion, so this comment does not apply anymore to this PR. I keep it for future reference.)

If pad is (0, 0) this is not executed.

pylearn2/models/mlp.py
fvisin: I have to admit I have no idea. It was part of the original code so I merged it, but it doesn't seem to be of any use. I will remove it.

pylearn2/models/mlp.py
fvisin: Great catch! :) 
Actually, does it even make sense to use `downsample.max_pool_2d` instead of `max_pool` as in the "non cudnn2d" case? Apparently there is no reason. I am not sure if the data will be in bc01 format, though I think that the data has been already reshaped in that form somewhere since the "normal" convolution does not force you to provide data in bc01 format. If you don't already know the answer, I will check if this is the case.

fvisin:
Thank you @lamblin and @nouiz for the super-fast support!
 
I removed the padding as I don't need it and it didn't seem to be handled very well.
Also, I removed the "special case" pooling falling back to the conv2d behavior that seems correct also in the cudnn case. This will allow to merge the other PR flawlessly. I am running an experiment on the new code, I will comment further if something goes wrong.

nouiz:
The other PR with change for max pooling was merged. Can you remove the change related to that that you have and rebase on master?

pylearn2/linear/cudnn2d.py
```diff
@@ -0,0 +1,162 @@
...
+        # x must be formatted as batch index, channel, topo dim 0, topo dim 1
+        # for use with conv2d, so check what the current input space format is
+        assert x.ndim == 4
+        axes = self.input_space.axes
+        assert len(axes) == 4
+
+        op_axes = ('b', 'c', 0, 1)
+
+        if tuple(axes) != op_axes:
+            x = x.dimshuffle(
```
JesseLivezey: I think you could write this as
```
x = x.dimshuffle(*[axes.index(ax) for ax in op_axes])
```
and it would be more concise and reuse the ```op_axes``` definition.

pylearn2/linear/cudnn2d.py
```diff
@@ -0,0 +1,162 @@
...
+        kerns = gpu_contiguous(self._filters)
+        desc = self.desc(img.shape, kerns.shape)
+        rval = self.conv_op(img, kerns, desc)
+
+        # Format the output based on the output space
+        axes = self.output_axes
+        assert len(axes) == 4
+
+        if tuple(axes) != op_axes:
+            rval = rval.dimshuffle(
```
JesseLivezey: Also here.

pylearn2/models/mlp.py
fvisin: I removed the padding.

pylearn2/linear/cudnn2d.py
fvisin: Yes, you are right. This commit was a very preliminary release, it needs refactoring. Thank you for pointing that out :)

pylearn2/linear/cudnn2d.py
```diff
@@ -0,0 +1,203 @@
...
+
+class Cudnn2D(OrigConv2D):
+    """
+    Wrapper on the Theano Cudnn op.
+
+    Parameters
+    ----------
+    filters : Theano shared variable
+        4D-tensor of shape (out channels, in channels, rows, cols)
+    batch_size : int
```
fvisin: Do we need batch_size? Can we get it from the input space?

pylearn2/linear/cudnn2d.py
```diff
@@ -0,0 +1,203 @@
...
+        The size of the input batches
+    input_space : Space
+        The Space of the input data
+    output_axes : tuple, optional
+        The requested output axes. If not specified `bc01` will be used.
```
fvisin: Does it make sense to default to bc01? Wouldn't it be better to default to the same axes as the input?

pylearn2/linear/cudnn2d.py
lamblin: The batch size is not part of the space. For the original Conv2D, batch_size could be None, indicating that the batch size was not constant. Can you check if this is supported by cudnn as well?

pylearn2/linear/cudnn2d.py
```diff
@@ -0,0 +1,195 @@
...
+    @functools.wraps(P2LT.get_weights_topo)
+    def get_weights_topo(self, borrow):
+        """
+        Parameters
+        ----------
+        borrow : TODO
+            TODO
+        """
+        return np.transpose(self._filters.get_value(borrow=borrow),
+                            (0, 2, 3, 1))
```
fvisin: This seems to assume filters is provided with a particular axes order. Can you confirm?
In this case I guess it would be better to generalize to whatever input axes where provided.

pylearn2/linear/cudnn2d.py
```diff
@@ -0,0 +1,195 @@
...
+        .. todo::
+
+            WRITEME
+
+        Parameters
+        ----------
+        batch_size : TODO
+            TODO
+        """
+        self._img_shape = tuple([batch_size] + list(self._img_shape[1:]))
```
fvisin: Why is `set_batch_size` modifying `self._img_shape`? And also, why is it not modifying `self._batch_size`?

pylearn2/linear/cudnn2d.py
lamblin: bc01 is the ordering of the output of the cudnn conv op.
Currently, pylearn2.linear.conv2d.Conv2D uses b01c as default (even if the Theano conv2d Op it uses outputs bc01), and pylearn2.linear.conv2d_c01b uses c01b (which is also the output of the op).
Using the minimal amount of dimshuffles makes sense to me at the moment, so the native format of the Op would be OK for me.

pylearn2/linear/cudnn2d.py
lamblin: There is no mechanism to provide or use axes for filters, so they need to be in the proper shape.
I don't think we want to introduce such a mechanism.

pylearn2/linear/cudnn2d.py
fvisin: Shuffle should be performed in place by numpy. Actually if I am not wrong, it should not even modify the array in memory, but directly index it in the right way when you access the data. 
Anyway, if you still think it is better to default to bc01 I'm ok with that.

pylearn2/linear/cudnn2d.py
lamblin: `self._img_shape` contains the shape for the whole input tensor (bc01).
There is no `self._batch_size`.

pylearn2/linear/cudnn2d.py
lamblin: I don't think it matters much, as the output_axes are currently always provided explicitly.
It is true that dimshuffles are performed inplace with almost no overhead, and chained dimshuffles would even be fused by Theano. However, there is a slightly higher risk that a "gpu_contiguous" gets inserted between dimshuffles, which would force a copy.
I would keep bc01, but I would not bother making the same change in conv2d.Conv2D.

pylearn2/linear/cudnn2d.py
fvisin: Uhm....cudnn is this file, I am confused. :) 

batch_size is only used to compute the img_shape parameter passed to the _original Conv2D_ (`pylearn2.packaged_dependencies.theano_linear.conv2d`) constructor. The exact same thing is done in `pylearn2.linear.conv2d`.

pylearn2/linear/cudnn2d.py
ballasn: cudnn does call gpu_contiguous, so I think it is important to keep the format bc01 to avoid unecessary copy

fvisin:
@nouiz I rebased on top of current master, thank you for notifying me that the other PR was merged.

pylearn2/linear/cudnn2d.py
lamblin: I mean, by the underlying Theano Op, GpuDnnConv.

pylearn2/linear/cudnn2d.py
fvisin: Ok. Thanks for pointing that out.

pylearn2/linear/cudnn2d.py
fvisin: Ok, I forgot the axes order is specified in the docstring so it's ok as is.

pylearn2/linear/cudnn2d.py
fvisin: The original `Conv2D` uses `batch_size` to define `image_shape`, that is then used as an argument for the constructors of both the `theano.tensor.nnet.conv` function and the `pylearn2.packaged_dependencies.theano_linear.conv2d` function.

`GpuDnnConv` seems not to use the batch_size at all. The underlying Theano Op [[doc]](http://deeplearning.net/software/theano/library/sandbox/cuda/dnn.html) [[code]](https://github.com/Theano/Theano/blob/master/theano/sandbox/cuda/dnn.py) takes only 3 parameters: `image`, `kernel`, `descr`.

Does this answer to your question?

lamblin:
The last thing missing for a merge is a unit test.

JesseLivezey:
I'm curious, since theano could already insert cudnn convolutions into the graph, is the main purpose of this PR to expose the additional functionality of the cudnn op compared to the conv2d op?

fvisin:
The reason for this PR is that Theano, at the moment at least, uses conv3D instead of cudnn if you convolve with stride greater than 1, which is highly inefficient.

lamblin:
Theano currently fails to insert cudnn convolution in once case: for the gradient of convolutions with subsampling. In that case, the gradient is computed by CPU-only conv3D.

pylearn2/linear/cudnn2d.py
```diff
@@ -0,0 +1,204 @@
...
+        op_axes = ('b', 'c', 0, 1)
+
+        if tuple(axes) != op_axes:
+            x = x.dimshuffle(*[axes.index(ax) for ax in op_axes])
+
+        # The calling format has to be changed
+        img = gpu_contiguous(x)
+        kerns = gpu_contiguous(self._filters)
+        desc = self._desc(img.shape, kerns.shape)
+        rval = self._conv_op(img, kerns, desc)
```
fvisin: This is not working as it is an implicit call to https://github.com/Theano/Theano/blob/fb4d5b4d50be1cda5ee4c49e9d4b88f0f2f817f3/theano/sandbox/cuda/dnn.py#L419 which needs another parameter (`output`). Can you suggest me a proper way to set `output`?

lamblin: If I remember correctly, "output" is a tensor into which the convolution will be written or accumulated.
You can call `gpu_alloc_empty` with the output shape to create it.

fvisin:
Ok, it should be finally done! Let's hope that Travis doesn't complain.
I recreated the same test as for conv2D, excluded those that didn't apply. Some of them are slightly different, but in the overall they should be pretty much the same.

pylearn2/linear/tests/test_cudnn.py
```diff
@@ -0,0 +1,151 @@
...
+
+import unittest
+from nose.plugins.skip import SkipTest
+
+import numpy as np
+
+skip_if_no_gpu()
+
+if cuda.cuda_available is False:
+    raise SkipTest('Optional package cuda disabled')
```
lamblin: Instead of these last 4 lines, it would be better to call `skip_if_no_gpu()` inside `TestCudnn.setUp`.
This should not interfere with the script checking the docstring format.

pylearn2/linear/cudnn2d.py
lamblin: (That was addressed.)

pylearn2/linear/tests/test_cudnn.py
```diff
@@ -0,0 +1,141 @@
...
+        Check whether the cudnn has stored the correct filters
+        """
+        self.assertTrue(np.all(
+            self.cudnn2d.get_weights_topo(borrow=True) ==
+            np.transpose(self.filters.get_value(borrow=True), (0, 2, 3, 1))))
+
+    def test_lmul(self):
+        """
+        Use SciPy's ndimage to check whether the convolution worked
+        correctly
```
lamblin: Can you update the docstring to say that we compare against conv2d, not scipy?

pylearn2/linear/tests/test_cudnn.py
```diff
@@ -0,0 +1,141 @@
...
+        """
+        Go from 2 to 3 channels and see whether the shape is correct
+        """
+        input_space = Conv2DSpace((3, 3), num_channels=3)
+        filters_values = np.ones(
+            (2, 3, 2, 2), dtype=theano.config.floatX
+        )
+        filters = sharedX(filters_values)
+        image = np.random.rand(1, 3, 3, 3).astype(theano.config.floatX)
+        conv2d = Conv2D(filters, 1, input_space)
```
lamblin: We should test Cudnn2D here instead, right?

pylearn2/linear/tests/test_cudnn.py
```diff
@@ -0,0 +1,141 @@
...
+                            conv2d.lmul(self.image_tensor))
+        assert f(image).shape == (1, 2, 2, 2)
+
+    def test_make_random_conv2D(self):
+        """
+        Create a random convolution and check whether the shape, axes and
+        input space are all what we expect
+        """
+        output_space = Conv2DSpace((2, 2), 1)
+        conv2d = make_random_conv2D(1, self.input_space, output_space,
```
lamblin: Idem here, this is testing the other conv2d.

pylearn2/models/mlp.py
```diff
@@ -22,11 +22,21 @@
...
+from theano.tensor.signal import downsample
+from theano.compat.six.moves import zip as izip
 
 from pylearn2.compat import OrderedDict
 from pylearn2.costs.mlp import Default
 from pylearn2.expr.probabilistic_max_pooling import max_pool_channels
-from pylearn2.linear import conv2d
+# Try to import the fast cudnn library, else fallback to conv2d
+cudnn2d = None
+if config.device == 'gpu':
```
lamblin: This will not work when users specify `'gpu0'` or `'gpu1'`, for instance.
A better test would be checking for `theano.sandbox.cuda.cuda_enabled` and `dnn_available`.

pylearn2/models/mlp.py
lamblin: OK.

pylearn2/models/mlp.py
```diff
@@ -2988,15 +2994,26 @@ def initialize_transformer(self, rng):
...
+                    rng=rng)
+            else:
+                self.transformer = conv2d.make_random_conv2D(
+                    irange=self.irange,
+                    input_space=self.input_space,
+                    output_space=self.detector_space,
+                    kernel_shape=self.kernel_shape,
+                    subsample=self.kernel_stride,
+                    border_mode=self.border_mode,
+                    rng=rng)
```
lamblin: If you detect this automatically, maybe it would be better to simply import `cudnn2d` as `conv2d`, and avoid that duplication here.
This would also ensure that `make_sparse_....` would use the cudnn version as well..

pylearn2/models/mlp.py
```diff
@@ -4455,11 +4473,6 @@ def set_mlp(self, mlp):
         super(FlattenerLayer, self).set_mlp(mlp)
         self.raw_layer.set_mlp(mlp)
 
-    @wraps(Layer.get_weights)
-    def get_weights(self):
-
-        return self.raw_layer.get_weights()
-
```
lamblin: Why did you remove that?

pylearn2/models/mlp.py
```diff
@@ -2961,14 +2967,14 @@ def __init__(self,
...
+                pool_type)
             assert pool_stride is not None, (
-                "You should specify the strides of "
-                "the spatial %s-pooling." % pool_type)
+                "You should specify the strides of the spatial %s-pooling." %
+                pool_type)
 
         assert nonlinearity is not None
-
+        super(ConvElemwise, self).__init__()
```
lamblin: Why move the call to `super` here?

pylearn2/models/mlp.py
fvisin: Because it's already defined here: https://github.com/fvisin/pylearn2/blob/55939adf102ecbeb3ac239be86e7960e36b25ad0/pylearn2/models/mlp.py#L756-L762

pylearn2/models/mlp.py
fvisin: If I am not wrong, it was like that in the code I was porting.
It seems reasonable to check the consistency of the parameters before calling the inherited constructor though, isn't it?

pylearn2/models/mlp.py
fvisin: Ok

pylearn2/models/mlp.py
fvisin: Ok.

pylearn2/linear/tests/test_cudnn.py
fvisin: Ops....sorry I didn't finish my job I guess!

pylearn2/linear/tests/test_cudnn.py
fvisin: Even worse. Sorry again.

pylearn2/linear/tests/test_cudnn.py
fvisin: Sure!

pylearn2/models/mlp.py
lamblin: Right.

pylearn2/models/mlp.py
lamblin: But that's not the same class, and not the same implementation

fvisin: Sorry I meant this line: https://github.com/fvisin/pylearn2/blob/55939adf102ecbeb3ac239be86e7960e36b25ad0/pylearn2/models/mlp.py#L4426-L4428

fvisin:
Ok, I am going to push the modification you requested. Let me know if I skip the test in the right way please.

pylearn2/models/mlp.py
lamblin: Right, OK then.

lamblin:
The way the tests are skipped look correct, I'll test on a machine without cudnn and let you know.
Regarding travis, there are still the following failures related to documentation:
```
======================================================================
FAIL: Test if pep8 is respected.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/miniconda/envs/pyenv/lib/python2.6/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/build/lisa-lab/pylearn2/pylearn2/devtools/tests/test_format.py", line 629, in test_format_pep8
    raise AssertionError("PEP8 Format not respected")
AssertionError: PEP8 Format not respected
-------------------- >> begin captured stdout << ---------------------
/home/travis/build/lisa-lab/pylearn2/pylearn2/linear/cudnn2d.py:259:1: W391 blank line at end of file
--------------------- >> end captured stdout << ----------------------
======================================================================
FAIL: Test if docstrings are well formatted.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/miniconda/envs/pyenv/lib/python2.6/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/build/lisa-lab/pylearn2/pylearn2/devtools/tests/test_format.py", line 664, in test_format_docstrings
    verify_format_docstrings()
  File "/home/travis/build/lisa-lab/pylearn2/pylearn2/devtools/tests/test_format.py", line 694, in verify_format_docstrings
    raise AssertionError("Docstring format not respected:\n%s" % msg)
AssertionError: Docstring format not respected:
/home/travis/build/lisa-lab/pylearn2/pylearn2/linear/tests/test_cudnn.py:
TestCudnn class has no Parameters section
/home/travis/build/lisa-lab/pylearn2/pylearn2/linear/cudnn2d.py:
make_sparse_random_conv2D:No Parameters section
```

pylearn2/linear/tests/test_cudnn.py
```diff
@@ -0,0 +1,146 @@
...
+    Tests for the Cudnn code
+    """
+    def setUp(self):
+        """
+        Set up a test image and filter to re-use
+        """
+        skip_if_no_gpu()
+        if not dnn_available():
+            SkipTest('Skipping tests cause cudnn is not available')
+        self.image = np.random.rand(1, 1, 3, 3).astype(theano.config.floatX)
```
lamblin: So I tried to run the tests, this fails when floatX is float64 (which is the default).
A solution would be to force floatX to be float32, as is done in https://github.com/fvisin/pylearn2/blob/mlp_cudnn/pylearn2/linear/tests/test_conv2d_c01b.py#L25, and restore that in `tearDown`. I don't think you have to worry about the call to `use` or `unuse`, though.

pylearn2/linear/tests/test_cudnn.py
```diff
@@ -0,0 +1,146 @@
...
+    """
+    Tests for the Cudnn code
+    """
+    def setUp(self):
+        """
+        Set up a test image and filter to re-use
+        """
+        skip_if_no_gpu()
+        if not dnn_available():
+            SkipTest('Skipping tests cause cudnn is not available')
```
lamblin: This should be `raise SkipTest(...)`, actually.

fvisin: Oh, right.

pylearn2/linear/tests/test_cudnn.py
fvisin: Ok. Everything should be ok now.

pylearn2/linear/tests/test_cudnn.py
```diff
@@ -0,0 +1,151 @@
...
+class TestCudnn(unittest.TestCase):
+    """
+    Tests for the Cudnn code
+    """
+    def setUp(self):
+        """
+        Set up a test image and filter to re-use
+        """
+        self.orig_floatX = theano.config.floatX
+        theano.config.floatX = 'float32'
```
lamblin: This needs to be done after the "SkipTest" exceptions can be raised, otherwise `tearDown` is not called, and the value is not restored.

pylearn2/linear/cudnn2d.py
```diff
@@ -0,0 +1,259 @@
...
+        filters=W,
+        batch_size=batch_size,
+        input_space=input_space,
+        output_axes=output_space.axes,
+        subsample=subsample,
+        border_mode=border_mode,
+        filters_shape=W.get_value(borrow=True).shape,
+        message=message
+    )
+
```
lamblin: Pep8 says: `W391 blank line at end of file`

pylearn2/linear/cudnn2d.py
```diff
@@ -0,0 +1,259 @@
...
+                              batch_size,
+                              subsample=(1, 1),
+                              border_mode='valid',
+                              message="",
+                              rng=None):
+    """
+    .. todo::
+
+        WRITEME properly
+
```
lamblin: Missing "Parameters" section.

pylearn2/linear/tests/test_cudnn.py
fvisin: Right, my bad.

pylearn2/linear/cudnn2d.py
fvisin: Sorry, I am messing with my vim settings and it's not behaving as I expect. I will fix it.

pylearn2/linear/cudnn2d.py
fvisin: I added it, but I am not sure on the meaning of the parameters so I left them empty.

pylearn2/linear/cudnn2d.py
lamblin: OK.

pylearn2/linear/cudnn2d.py
```diff
@@ -0,0 +1,269 @@
...
+    Creates a Cudnn2D with random kernels, where the randomly initialized
+    values are sparse
+
+    Parameter
+    ---------
```
lamblin: The section should be named "Parameters".
Rather than have empty descriptions, can you put "TODO"?

pylearn2/linear/tests/test_cudnn.py
```diff
@@ -0,0 +1,151 @@
...
+import unittest
+from nose.plugins.skip import SkipTest
+
+import numpy as np
+
+
+class TestCudnn(unittest.TestCase):
+    """
+    Tests for the Cudnn code
+    """
```
lamblin: It's complaining about a missing "Parameters" section here, I'm not really sure why.
Maybe you can put a Parameters section that just says to refer to unittest.TestCase and check it fixes the error.

pylearn2/linear/tests/test_cudnn.py
```diff
@@ -0,0 +1,151 @@
...
+        self.input_space = Conv2DSpace((3, 3), 1, axes=('b', 'c', 0, 1))
+        self.filters_values = np.ones(
+            (1, 1, 2, 2), dtype=theano.config.floatX
+        )
+        self.filters = sharedX(self.filters_values, name='filters')
+        self.batch_size = 1
+
+        self.cudnn2d = Cudnn2D(self.filters, self.batch_size, self.input_space)
+
+    def tearDown(self):
```
lamblin: Missing docstring.

fvisin: Yep, fixed that. Apparently the latest pylint installation on ubuntu broke syntastic :/
I am pushing frequently to find out syntax/pep8 errors with Travis, don't bother pointing them out. I will check Travis logs on my own and I'll comment back when I think I am done, so you don't waste your time on stupid errors like this. :)

fvisin:
Finally Travis is happy with it! I ran the tests on GPU with cudnn and they all pass, I didn't check with a GPU that doesn't support cudnn though, if you want to try that again.

lamblin:
@fvisin Thanks! I'll test on a GPU without cudnn, and merge if it passes.

pylearn2/linear/tests/test_cudnn.py
```diff
@@ -0,0 +1,168 @@
...
+    Refer to unittest.TestCase.
+
+    """
+    def setUp(self):
+        """
+        Set up a test image and filter to re-use.
+        """
+        skip_if_no_gpu()
+        if not dnn_available():
+            raise self.SkipTest('Skipping tests cause cudnn is not available')
```
lamblin: It is just `SkipTest`, not `self.SkipTest`.

pylearn2/linear/tests/test_cudnn.py
lamblin: OK.

lamblin:
OK, so the last missing thing is to fix the syntax of `SkipTest`, otherwise the tests fail on a machine with a GPU but not cudnn.

fvisin:
Fixed! Should be possible to merge now.

Francesco

lamblin:
Thanks a lot!

jstaker7:
Thanks for the fix! I had to change this line in mlp.py:
From:
from linear import cudnn2d as conv2d
To:
from pylearn2.linear import cudnn2d as conv2d

Because linear had not been referenced yet.

fvisin:
Thank you @jstaker7 for pointing it out. Should be fixed soon by https://github.com/lisa-lab/pylearn2/pull/1507

Issue #1408, opened 2015-02-16
Deprecate sparse auto encoders


nouiz:
It is broken for at least 2 years...

PR #1409, opened 2015-02-16
deprecated SparseAutoencoder, code deleted

yaoli:
Following Fred's suggestion, I officially deprecated this class, deleting all the historical codes.

fix gh-1408


bartvm:
You should raise the error in __init__ so that no work is done on import (which is crashing the documentation generator).

nouiz:
As this is very old stuff that do not work, I suggested a faster deprecation, as having code crash for years should count as deprecation I think.

I'm good with this PR. Is there someone having a problem with merging this? I'll wait about 1 day before merging this PR.

PR #1411, opened 2015-02-20
OneHotFormatter: include dtype in symbolic expression

hans:
Simple fix: symbolic expressions for OneHotFormatter currently default to float32 (the formatter's dtype isn't passed on).

Thanks for the great library!

### DIFF
--- a/pylearn2/format/target_format.py
+++ b/pylearn2/format/target_format.py
@@ -201,7 +201,7 @@ class OneHotFormatter(object):
         else:
             if mode == 'concatenate':
                 one_hot = tensor.zeros((targets.shape[0] * targets.shape[1],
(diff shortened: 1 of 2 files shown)


lamblin:
Thanks for the fix!
Could you add a small unit test in `pylearn2/format/tests/test_target_format.py` so we can check it does not get broken again in the future?

hans:
I added a simple `dtype` check to the existing simple and symbolic formatter tests.

lamblin:
I'm not sure how adding assertions in a test managed to make a different test fail.
I'll trigger a new merge, and see if it helps.

lamblin:
Thanks for the PR!

PR #1414, opened 2015-02-24
max_pool works with anon variable

JesseLivezey:
Added test failed before change.

https://groups.google.com/forum/#!topic/pylearn-dev/RxGwmSvuqtg


lamblin:
Thanks for the fix, sorry about the delay.

PR #1415, opened 2015-02-25
Fix various tests on the daily buildbot

lamblin:


nouiz:
@goodfeli @dwf 
Can you check this PR? It fix problems in the daily buildbot, it would be great to clean it as much as possible before the Theano release.

From my point of view, this PR is ok, but there is stuff I can't do a good review.

goodfeli:
did this get rid of the labels for TL challenge?

lamblin:
@goodfeli No, the labels are in `self.y_fine` and `self.y_coarse`, they were not in `self.y`, and were not accessible via the usual `DenseDesignMatrix` interface to begin with.

PR #1063, opened 2014-08-06
ArangeDataset now inherits the parameters defined by its superclass

AlassaneNdiaye:
I think this could be useful if we want to test other features of Datasets such as preprocessed Datasets.

PR #1419, opened 2015-02-27
Update arange dataset

lamblin:
This is a rebase of #1063


lamblin:
Travis passes, and this is just a rebase of Alassane's commits, so I'm merging.

PR #1420, opened 2015-02-27
Catch exception in the local cache and handle that case.

nouiz:


pylearn2/datasets/cache.py
```diff
@@ -116,7 +116,14 @@ def cache_file(self, filename):
 
         # Create the folder structure to receive the remote file
         local_folder = os.path.split(local_name)[0]
-        self.safe_mkdir(local_folder)
+        try:
+            self.safe_mkdir(local_folder)
+        except Exception, e:
```
lamblin: This syntax does not work for Python 3, I think it should be
    except Exception as e:

nouiz:
I didn't know Pylearn2 passed to a single code base for python2 and 3. I amended the commit.

lamblin:
Thanks!

PR #1423, opened 2015-03-04
removed bad check, fixes #1338

### LINKED ISSUE #1338: CompositeLayer fails if number of inputs is larger than number of layers
In my MLP model I have a composite space with two components as input:

    input_space: !obj:pylearn2.space.CompositeSpace {
      components: [
        !obj:pylearn2.space.VectorSpace {
          dim: 50, 
        },
        !obj:pylearn2.space.IndexSpace {
          dim: 1, 
          max_labels: 100
        }
      ]
    },
    input_source: ['features', 'additional_stuff'']

In my model, I want to discard the second component, so I use a CompositeLayer with only one layer, and specify that input 0 is routed to layer 0 and input 1 to nothing:

    !obj:pylearn2.models.mlp.CompositeLayer {
          layer_name: 'composite_layer',
          inputs_to_layers: {0: [0], 1: []},
          layers: [
            !obj:pylearn2.models.mlp.RectifiedLinear {
                           layer_name: 'h0',
                           dim: 300,
                           sparse_init: 15,
                           max_col_norm: 5.0,
            },
          ],
        },

This causes an assert at https://github.com/lisa-lab/pylearn2/blob/master/pylearn2/models/mlp.py#L4071 to fail. The assertion is invalid, as it checks, whether the ID of an *input* is less than the number of *layers*. 

I think the number of actual inputs is not known at that point, so I would recommend to just remove that line (works according to my experiments).

JesseLivezey:
Test failed before the line was removed. The correct check is already done in CompositeLayer.set_input_space() here:
https://github.com/lisa-lab/pylearn2/blob/e9338af5050b7dda2fa87204d4eee89a3ed8fbdb/pylearn2/models/mlp.py#L4165

### DIFF
--- a/pylearn2/models/mlp.py
+++ b/pylearn2/models/mlp.py
@@ -4135,7 +4135,6 @@ class CompositeLayer(Layer):
             self.inputs_to_layers = OrderedDict()
             for key in sorted(inputs_to_layers):
                 assert isinstance(key, py_integer_types)
-                assert 0 <= key < self.num_layers
                 value = inputs_to_layers[key]
                 assert is_iterable(value)
                 assert all(isinstance(v, py_integer_types) for v in value)
--- a/pylearn2/models/tests/test_mlp.py
+++ b/pylearn2/models/tests/test_mlp.py
@@ -329,6 +329,42 @@ def test_multiple_inputs():
     train.main_loop()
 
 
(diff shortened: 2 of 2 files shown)


lamblin:
Thanks!

PR #1425, opened 2015-03-04
[WIP] Adam learning rule

JesseLivezey:
This implementation is based on the arxiv [v4] paper. Haven't run or tested it yet.

The paper seems to have at least one typo in that \beta_2^t is used but never defined. I'm assuming it is just \beta_2 currently. Also assuming that \beta_{1,t} is the same thing as \beta_1^t.

### DIFF
--- a/pylearn2/training_algorithms/learning_rule.py
+++ b/pylearn2/training_algorithms/learning_rule.py
@@ -502,8 +502,9 @@ class RMSProp(LearningRule):
 
 class Adam(LearningRule):
     """
-    Implements the AdaDelta learning rule as described in:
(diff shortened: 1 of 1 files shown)


goodfeli:
Why not just use Alec Radford's implementation?
https://gist.github.com/Newmu/acb738767acb4788bac3

I've been using that plugged into Pylearn2 in my private repo and it works well.

JesseLivezey:
I don't think Alec's version is consistent with the most recent version of the paper, but I haven't really tested this implementation vs. his, so I'm not sure how different the results will be.

JesseLivezey:
It might just be that Alec's version doesn't decay beta1, although the betas have been redefined, and I haven't checkout to see whether the rest of the math is equivalent.

PR #1426, opened 2015-03-09
Small fixes

### LINKED ISSUE #1157: Fix devtools/tests/test_format.py and errors it detects
This file is executable, which makes nosetests skip it by default.
We should:
- change the permissions
- fix all problems detected by `test_format_pep8()` and `test_format_docstrings()` in that file.

### LINKED ISSUE #1354: compute_test_err.py is broken for mnist.yaml in maxout
This line:
https://github.com/lisa-lab/pylearn2/blob/master/pylearn2/scripts/papers/maxout/compute_test_err.py#L55
will fail because x_arg will already be in a Conv2D space for mnist.yaml.

See discussion here:
https://groups.google.com/forum/#!topic/pylearn-users/WAo8x3LzvdU

lamblin:
First commit fixes #1354.
Second commit fixes #1157

### DIFF
--- a/pylearn2/scripts/papers/maxout/compute_test_err.py
+++ b/pylearn2/scripts/papers/maxout/compute_test_err.py
@@ -51,8 +51,6 @@ def accs():
                             data_specs = model.cost_from_X_data_specs())
     for item in iterator:
         x_arg, y_arg = item
-        if Xb.ndim > 2:
-            x_arg = test.get_topological_view(x_arg)
         mf1_accs.append(batch_acc(x_arg, y_arg)[0])
     return sum(mf1_accs) / float(len(mf1_accs))
 
--- a/doc/internal/pull_request_checklist.txt
+++ b/doc/internal/pull_request_checklist.txt
@@ -325,3 +325,8 @@ Use the `with` statement, i.e.::
 
 This is cleaner and ensures that the file always gets closed, even in an
 error condition.
+
+Are you adding new files or changing files permissions?
+=======================================================
+The files containing unit tests (named `test_....py`) should never be
+executable, otherwise nose will ignore them, and not execute the tests.


pylearn2/scripts/papers/maxout/compute_test_err.py
```diff
@@ -51,8 +51,6 @@ def accs():
                             data_specs = model.cost_from_X_data_specs())
     for item in iterator:
         x_arg, y_arg = item
-        if Xb.ndim > 2:
-            x_arg = test.get_topological_view(x_arg)
```
bouthilx: Was it because x_arg used to be 2D no matter what ```model.get_input_space().make_batch_theano()``` shape was? Where is the conversion done in the iterator?

bouthilx:
@lamblin Did you fix the executable rights to test*.py files? There is none in the commit.

pylearn2/scripts/papers/maxout/compute_test_err.py
lamblin: @bouthilx Actually, in the past, the iterator was not used, and the script looped directly on `test.X`: https://github.com/lisa-lab/pylearn2/commit/ff954c7599f83157d9caba13bf39989ae7ae803c
`FiniteDatasetIterator` will create the appropriate `convert` functions for the sources where the `Space` requested is different from the "native" one: https://github.com/lisa-lab/pylearn2/blob/master/pylearn2/utils/iteration.py#L926
These functions will then be called on each minibatch by the iterator's `next()` method.

lamblin:
The executable rights had been fixed a few months ago: https://github.com/lisa-lab/pylearn2/pull/1159

pylearn2/scripts/papers/maxout/compute_test_err.py
bouthilx: Ah, OK I found it in DenseDesignMatrix. Thanks!

PR #1429, opened 2015-03-11
Added X_labels and y_labels to DenseDesignMatrixPyTables constructor.

### LINKED ISSUE #1294: Make DenseDesignMatrixPyTables use x_labels and y_labels
Its constructor should accept keywords for its parent contructor.

eivind88:
Fixes #1294


lamblin:
Thanks for the pull request.
Could you add a unit test that checks this works as expected when a DenseDesignMatrixPyTables is built with those arguments?
Also, you will have to redefine `_check_labels`, because the definition in the base class does the following:
```
assert np.all(self.X < self.X_labels)
...
assert np.all(self.y < self.y_labels)
```
which is not something you want to do on the whole dataset.

eivind88:
Is this OK?

lamblin:
Thanks!

PR #1431, opened 2015-03-12
Ipython embed script

sankar-mukherjee:


lamblin:
The state of this PR is strange:
- it cannot be merged because of conflicts
- it only contains commits by someone else
My guess is that you tried a rebase that did not work as intended.
If you still have your original changes in one commit, can you try to rebase on top of the latest master?
Thanks.

PR #1432, opened 2015-03-12
Fix some other tests on the daily buildbot

lamblin:


lamblin:
This has to be merged  before https://github.com/Theano/Theano/pull/2611 gets accepted, as it removes `tensor.shared`

PR #1433, opened 2015-03-13
ConvElemwise now sets tied_b to the correct default

### LINKED ISSUE #1430: tied_b initialized to None instead of True in ConvElemWise
tied_b should be initialized to True according to the documentation. However, it is initialized to None, essentially resulting in a default value of False. Either the documentation or the default value should be updated.

GeertLitjens:
This commit should fix #1430. One thing I noticed is that the default is overwritten in ConvRectifiedLinear, should this not be True there also?

### DIFF
--- a/pylearn2/models/mlp.py
+++ b/pylearn2/models/mlp.py
@@ -2969,6 +2969,9 @@ class ConvElemwise(Layer):
 
         assert nonlinearity is not None
 
+        if tied_b is None:
+            self.tied_b = True
+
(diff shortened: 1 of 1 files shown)


lamblin:
Actually, this would change the default behaviour without a warning in advance, and I'm not sure whether the behaviour or the documentation should be changed (especially if the default of conv rectifiers is False).
Could you open a thread on pylearn-dev to discuss the issue?

PR #1439, opened 2015-03-17
fix SequenceDatasetIterator's small bug

ronekko:
`batch` might have a shape like (batch_size, sequence_length, feature, feature, ...).
The original code uses `np.transpose(batch, (1, 0, 2))` to swap the 0th and the 1st axes.
But using `np.swapaxis(batch, 0, 1)` is better than `np.transpose`.

Currently this is not a problem since the possible dimension of `batch` is only 3.
But in the future, if one want to use the space `SequenceDataSpace(Conv2DSpace)` (currently this is not allowed to use with RNN), the shape of `batch` will be (batch_size, sequence_length, row, col, channel) which dimension is 5 and `np.transpose(batch, (1, 0, 2))` will not match its shape.

### DIFF
--- a/pylearn2/sandbox/rnn/utils/iteration.py
+++ b/pylearn2/sandbox/rnn/utils/iteration.py
@@ -126,7 +126,7 @@ class SequenceDatasetIterator(FiniteDatasetIterator):
                 # Create mask
                 if source in self.mask_needed:
                     mask = self._create_mask(rval)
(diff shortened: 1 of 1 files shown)


lamblin:
Since it does not break anything for the moment, and makes it slightly easier to generalize that in the future (although right now `_create_mask` assumes a `VectorSpace`), I don't see any harm in that, so I'll merge, even though the utility is quite limited.

PR #1440, opened 2015-03-18
add kwargs to __init__ of CSVData

junbochen:
add kwargs to __init__() of CSVData, so that this class could use features of it's parent class, DenseDesignMatrix.
For example, the view_converter of  DenseDesignMatrix could not be initialized without this change.


lamblin:
Thanks!

PR #1441, opened 2015-03-18
fix incorrect parameter name in docstrings of LSTM and GRU

ronekko:
The docstrings of `LSTM` and `GRU` about `output` parameter are incorrect. 
Actual parameter name that they expect is `indices` as `Recurrent`.

The docstring of `Recurrent` class has been correctly fixed at [this commit](https://github.com/lisa-lab/pylearn2/commit/df3a0c54af23431042c30063de4615b916cf6c89#diff-99dd2c7f58dcebb88cf63a031a0a017aR37). 
But at that time the `LSTM`'s docstring was left incorrect.
After that, the `GRU`'s docstring is copied from `LSTM`'s one at [this commit](https://github.com/lisa-lab/pylearn2/commit/62bc2cc3689196cbe38346bf389188a604f8b44e#diff-99dd2c7f58dcebb88cf63a031a0a017aR620).

### DIFF
--- a/pylearn2/sandbox/rnn/models/rnn.py
+++ b/pylearn2/sandbox/rnn/models/rnn.py
@@ -422,12 +422,13 @@ class LSTM(Recurrent):
         The name of the layer. All layers in an MLP must have a unique name.
     irange : float
         Initializes each weight randomly in U(-irange, irange)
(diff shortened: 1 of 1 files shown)


lamblin:
Thanks!

PR #1442, opened 2015-03-18
Add corruptor in DeepComposedAutoencoder, fix #1356

### LINKED ISSUE #1356: Monitor stores num_examples as a float, causing failures with large numbers of examples
```
  File "/home/ubuntu/pylearn2/pylearn2/train.py", line 196, in main_loop
    self.run_callbacks_and_monitoring()
  File "/home/ubuntu/pylearn2/pylearn2/train.py", line 242, in run_callbacks_and_monitoring
    self.model.monitor()
  File "/home/ubuntu/pylearn2/pylearn2/monitor.py", line 264, in __call__
    (ne, actual_ne))
RuntimeError: At compile time, your iterator said it had 60002992 examples total, but at runtime it gave us 60002994.
```

This seems to be caused by monitor.py line 441-442:
```
        self.num_examples = [np.cast[config.floatX](float(i.num_examples))
                             for i in it]
```

a float32 can't represent the number 60002994:
```
>>> import numpy as np
>>> np.cast["float32"](float(60002994.0))
array(60002992.0, dtype=float32)
```

woozzu:
1. Add corruptor in DeepComposedAutoencoder, for implementing deep denoising autoencoder
2. Fix #1356

### DIFF
--- a/pylearn2/models/autoencoder.py
+++ b/pylearn2/models/autoencoder.py
@@ -716,7 +716,7 @@ class DeepComposedAutoencoder(AbstractAutoencoder):
     autoencoders : list
         A list of autoencoder objects.
     """
-    def __init__(self, autoencoders, corruptor=None):
+    def __init__(self, autoencoders):
         super(DeepComposedAutoencoder, self).__init__()
         self.fn = None
         self.cpu_only = False
@@ -727,21 +727,6 @@ class DeepComposedAutoencoder(AbstractAutoencoder):
         self.autoencoders = list(autoencoders)
         self.input_space = autoencoders[0].get_input_space()
         self.output_space = autoencoders[-1].get_output_space()
-        self.corruptor = corruptor
-
-    @functools.wraps(Autoencoder.reconstruct)
-    def reconstruct(self, inputs):
-        """
-        .. todo::
-
-            WRITEME
-        """
-        corrupted = None
-        if self.corruptor is None:
-            corrupted = inputs
-        else:
-            corrupted = self.corruptor(inputs)
-        return self.decode(self.encode(corrupted))
 
     @functools.wraps(Autoencoder.encode)
     def encode(self, inputs):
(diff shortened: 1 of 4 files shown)


lamblin:
1. That change makes all DeepComposedAutoencoders into denoising autoencoders. I think it would be clearer to have a new class for that, inheriting from DeepComposedAutoencoders (and maybe DenoisingAutoEncoder as well)
2. Two things are still missing: unit tests, and make sure that monitors serialized previously can still be loaded.

pylearn2/models/autoencoder.py
```diff
@@ -727,6 +727,21 @@ def __init__(self, autoencoders):
...
         self.output_space = autoencoders[-1].get_output_space()
+        self.corruptor = corruptor
+
+    @functools.wraps(Autoencoder.reconstruct)
+    def reconstruct(self, inputs):
+        """
+        .. todo::
+
+            WRITEME
+        """
```
lamblin: You should not add new "WRITEME" docstrings. Especially in that case, since if you use `wraps`, the docstring of the wrapped method will be used.

pylearn2/models/autoencoder.py
```diff
@@ -716,7 +716,7 @@ class DeepComposedAutoencoder(AbstractAutoencoder):
     autoencoders : list
         A list of autoencoder objects.
     """
-    def __init__(self, autoencoders):
+    def __init__(self, autoencoders, corruptor=None):
```
lamblin: If you add a new parameter to a constructor, add it to the class's docstring.

pylearn2/monitor.py
```diff
@@ -474,8 +473,8 @@ def redo_theano(self):
                 if n == 0:
                     raise ValueError("Iterating over 0 examples results in " +
                                      "divide by 0")
-                val = (channel.val * T.cast(batch_size, config.floatX)
-                       / cur_num_examples)
+                val = T.cast(channel.val * batch_size / cur_num_examples,
```
lamblin: On python2, this may be integer division, which is not what you want.

woozzu:
@lamblin Thank you for your comment. How can I verify that old serialized monitor can be loaded correctly? pkl file is in ignore list.

woozzu:
@lamblin Thank you for your comment.
The only change of monitor is the type of self.num_examples. In my observation, this variable is not serialized (I could not find the variable when I loaded a serialized monitor)

woozzu:
@lamblin And also I am not sure if inheriting from DenoisingAutoencoder is appropriate because DeepComposedAutoencoder(including denoising one) does not use variables such as nvids, nhids and so on.

lamblin:
Regarding `StackedDenoisingAutoencoder`, I agree it does not make sense to inherit from `DenoisingAutoencoder`, and your change looks reasonable.

Also, in the future, it is better to keep unrelated changes (autoencoder, and dtype of num_examples) in separate pull requests.

woozzu:
@lamblin Thank you a lot for your advice. I will keep it in mind.

lamblin:
Other than that, I'm good to merge.

PR #1451, opened 2015-03-24
More general check in data_specs

### LINKED ISSUE #1450: Improving mapping.nest error message
If you pass mapping.nest something that is a list rather than a tuple, you get a completely unrelated error message:

      File "/usr/local/google/home/goodfellow/pylearn2/pylearn2/utils/data_specs.py", line 273, in nest
        assert self.n_unique_specs == 1
    AssertionError

bartvm:
Fixes #1450.

### DIFF
--- a/pylearn2/utils/data_specs.py
+++ b/pylearn2/utils/data_specs.py
@@ -3,6 +3,7 @@ Utilities for working with data format specifications.
 
 See :ref:`data_specs` for a high level overview of the relevant concepts.
 """
+from collections import Sized
 from pylearn2.space import CompositeSpace, NullSpace, Space
 from pylearn2.utils import safe_zip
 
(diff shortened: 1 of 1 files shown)


pylearn2/utils/data_specs.py
```diff
@@ -265,7 +266,7 @@ def nest(self, flat):
                 assert self.n_unique_specs == 1
             return self._make_nested_space(flat, self.spec_mapping)
         else:
-            if isinstance(flat, tuple):
+            if isinstance(flat, Sized):
```
lamblin: The problem with "Sized" is that it would match a string, and try to match the length of the string with the number of elements needed, whereas in that case we want `n_unique_specs` to be 1, and the whole string to be nested.

bartvm: And just checking for `(tuple, list)`, would that be better?

lamblin: Yes, but you would need to convert the list to a tuple before calling `_make_nested_tuple`, or update `_make_nested_tuple` in the same way.

bartvm: I haven't read the code in detail, but I guess that changing all the `isinstance` checks is better? Just in case another method or the user decides to call these methods without converting, which would just re-introduce the bug.

PR #1455, opened 2015-03-25
fixes #1454 (isinstance instead of is_numeric in space submodule)

### LINKED ISSUE #1454: [bug] _validate_impl for Conv2DSpace
[This line](https://github.com/lisa-lab/pylearn2/blob/master/pylearn2/space/__init__.py#L1968):
```python
        if isinstance(batch, theano.gof.Variable):
```
should be
```python
        if not is_numeric:
```
right?

TNick:


lamblin:
Thanks!

PR #1456, opened 2015-03-26
Allowing modification of learning rate for mu

XavierAudier:
In Gaussian visible layer, mu should be able to have a learning rate different from 1/sharing.


lamblin:
Thanks.
Can you add (or expand existing) unit tests to check that the options are correctly recognized and the behaviour is as expected?

XavierAudier:
I've added a test for learning rates : #1460. It makes sure the learning rate for mu is implemented in the same way it is for beta. It runs fine on a dataset but I haven't put that in the test.

PR #1457, opened 2015-03-26
Added test for ZCA_Dataset.

### LINKED ISSUE #899: Add unit tests for ZCA_Dataset
People keep breaking this. Every time I try to use it it's broken in a new way.

Thrandis:
Fixes #899

### DIFF
--- a/pylearn2/datasets/tests/test_zca_dataset.py
+++ b/pylearn2/datasets/tests/test_zca_dataset.py
@@ -25,15 +25,15 @@ def test_zca_dataset():
     rng = np.random.RandomState([2014, 11, 4])
     raw = random_dense_design_matrix(rng, x.shape[0], x.shape[1], 2)
     raw.X = x
-    zca = ZCA()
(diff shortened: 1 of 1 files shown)


Thrandis:
Well it was always failing otherwise.
I ran some tests using between 10 and 1000 examples, and the variance gets never close to 1, but rather around 0.9. Maybe am I using too few examples?

Thrandis:
Well it was always failing otherwise.
I ran some tests using between 10 and 1000 examples, and the variance gets never close to 1, but rather around 0.9. Maybe am I using too few examples?

dwf:
Perhaps try with `filter_bias=0` and see if that fixes it.

Thrandis:
Yep it works! The tolerance can be pushed to 0.01. Is it precise enough?

dwf:
Sounds good to me.

memimo:
Sorry for late review. LGTM too. I am merging.
@Thrandis for future for unit tests that needs some preparation you can use python's unit test class and divide each test in different class method.

PR #1459, opened 2015-03-27
Expand RNN to work with ConvRectifiedLinear layer

ronekko:
This PR expand `RNN` to work with `ConvRectifiedLinear` layer.

To do this, I made the family of `Recurrent` classes capable of having `ConvRectifiedLinear` as its upstream layer, based on the mechanism of `Linear` class which can adapt to connect with `ConvRectifiedLinear`.

The list of changes:
- To make `Recurrent.set_input_space` accept `Conv2DSpace` as unit space of sequence type spaces.
- Since `Recurrent` layer expects the input data format as `VectorSpace` in `fprop`, if the input data format is `Conv2DSpace` then it is converted to `VectorSpace` format.
- To make `SequenceDataSpace.format_as` capable of conversion from `SequenceDataSpace(Conv2DSpace)` to `SequenceDataSpace(VectorSpace)`
- To prevent `fprop` methods from **multiple-wrapping** by `RNNWrapper.__new__`

Let me explain why the multiple-wrapping is a problem.
In `fprop_wrapper.outer`, the data (`state_below`) is unpacked when `require_unmask` is True. If a `fprop` is wrapped more than once, the already unpacked data is unpacked again at the second time of the same line and it fails. Thus `fprop` should be wrapped just once.

Multiple-wrapping arises for the layer classes which does not override `fprop` of its super class and `ConvRectifiedLinear` class falls into this category. In the class definition of a layer, if a method is not overridden then a re-wrapped verson of super's methods is registered as subclass's own method by `type.__new__` (not inherited).

One solution to prevent the use of `fprop` with multiple-wrapping is just to use super's `fprop` (means inheritance). Therefore, if `dct` does not contain `fprop`, it is excluded from new `dct`.

Multiple-wrapping arises not only for `ConvRectifiedLinear.fprop` but also all being wrapped and non-overridden methods of subclass of `Layer` class. In fact, the case of `fprop` is problematic, but in another case such as `get_output_space` or `get_target_space`, multiple-wrapping is necessary. 

Although it is easy to remove multiple-wrapping from `RNNWrapper`, it will cause another problem as mentioned above. There is a little complicated issue related to the other part of the codebase. So in this PR, I only implemented a workaround for multiple-wrapping of `fprop`.

### DIFF
--- a/pylearn2/sandbox/rnn/models/rnn.py
+++ b/pylearn2/sandbox/rnn/models/rnn.py
@@ -228,7 +228,10 @@ class Recurrent(Layer):
         self.initialize_params()
 
     def initialize_params(self):
-        # Initialize the parameters
+        """
+        This method initializes the parameters of the class.
+        Re-running this function will reset the parameters.
+        """
         rng = self.mlp.rng
         if self.irange is None:
             raise ValueError("Recurrent layer requires an irange value in "
@@ -465,7 +468,10 @@ class LSTM(Recurrent):
         del self.self
 
     def initialize_params(self):
-        # Initialize the parameters
+        """
+        This method initializes the parameters of the class.
+        Re-running this function will reset the parameters.
+        """
         rng = self.mlp.rng
         if self.irange is None:
             raise ValueError("Recurrent layer requires an irange value in "
@@ -634,7 +640,10 @@ class GRU(Recurrent):
         del self.self
 
     def initialize_params(self):
-        # Initialize the parameters
+        """
+        This method initializes the parameters of the class.
+        Re-running this function will reset the parameters.
+        """
         rng = self.mlp.rng
         if self.irange is None:
             raise ValueError("Recurrent layer requires an irange value in "


ronekko:
This build error of Travis occured due to errors of Github...
The build log shows that the tests are aborted at the very beginning of the build (cloning the repository failed).
Could you rebuild to run this test again?

ronekko:
Sorry, I closed this PR by mistake so I reopened this again.

pylearn2/sandbox/rnn/models/mlp_hook.py
```diff
@@ -51,6 +51,9 @@ def __new__(cls, name, bases, dct):
                     if attr.endswith('_wrapper')]
         for wrapper in wrappers:
             if wrapper not in dct:
+                # avoid multiple wrapping for fprop
+                if wrapper == "fprop":
+                    continue
```
lamblin: I don't think this is necessary, because "fprop" does not end with '_wrapper'.

pylearn2/sandbox/rnn/space/__init__.py
```diff
@@ -116,11 +110,11 @@ def _format_as_impl(self, is_numeric, batch, space):
...
             if isinstance(space, SequenceDataSpace):
                 if is_numeric:
-                    formatted_batch = np.transpose(np.asarray([
+                    formatted_batch = np.swapaxes(np.asarray([
                         self.space._format_as_impl(is_numeric, sample,
                                                    space.space)
-                        for sample in np.transpose(batch, (1, 0, 2))
-                    ]), (1, 0, 2))
+                        for sample in np.swapaxes(batch, 0, 1)
+                    ]), 0, 1)
```
lamblin: I don't think it is that simple if `space.space` does not have `b` as first dimension, for instance if it is `Conv2DSpace(axes='c01b')`.

pylearn2/sandbox/rnn/space/__init__.py
```diff
@@ -69,16 +69,10 @@ def __str__(self):
...
         assert isinstance(space, SequenceSpace)
-        if is_numeric:
-            rval = np.apply_over_axes(
-                lambda batch, axis: self.space._format_as_impl(
-                    is_numeric=is_numeric,
-                    batch=batch,
-                    space=space.space),
-                batch, 0)
-        else:
-            NotImplementedError("Can't convert SequenceSpace Theano variables")
```
lamblin: I'm not confident in the implementation of `SequenceDataSpace._format_as_impl` with `is_numeric=False`, maybe it is why it was disabled here.

pylearn2/sandbox/rnn/space/__init__.py
```diff
@@ -116,11 +110,11 @@ def _format_as_impl(self, is_numeric, batch, space):
...
+                    formatted_batch = np.swapaxes(np.asarray([
                         self.space._format_as_impl(is_numeric, sample,
                                                    space.space)
-                        for sample in np.transpose(batch, (1, 0, 2))
-                    ]), (1, 0, 2))
+                        for sample in np.swapaxes(batch, 0, 1)
+                    ]), 0, 1)
                 else:
                     formatted_batch, _ = scan(
                         fn=lambda elem: self.space._format_as_impl(
```
lamblin: Here, I find it strange that the batch is not transposed, like it is in the `is_numeric` case.

pylearn2/sandbox/rnn/space/__init__.py
```diff
@@ -181,11 +175,16 @@ def get_origin_batch(self, batch_size, dtype=None):
 
     @wraps(space.Space.make_theano_batch)
     def make_theano_batch(self, name=None, dtype=None, batch_size=None):
-
         dtype = self._clean_dtype_arg(dtype)
-        broadcastable = [False] * 3
-        broadcastable[1] = (batch_size == 1)
-        broadcastable[2] = (self.dim == 1)
+        if not isinstance(self.space, Conv2DSpace):
```
lamblin: Here, it would be better to whitelist `VectorSpace` and maybe `IndexSpace`, rather than assuming that everything except `Conv2DSpace` has 2 dimensions with the first one being the batch.

pylearn2/sandbox/rnn/space/tests/test_sequence_space.py
```diff
@@ -0,0 +1,162 @@
...
+import theano
+from pylearn2.space import VectorSpace, Conv2DSpace
+from pylearn2.sandbox.rnn.space import SequenceSpace
+
+floatX = theano.config.floatX
+
+
+def test_sequence_vector_space_make_theano_batch():
+    """
+    SequenceSpate(VectorSpace).make_theano_batch should create 3D tensor
```
lamblin: "SequenceSpace" :)

lamblin:
Sorry for the delay, this is sandbox code that is not well-known or maintained.
@bartvm would you like to have a look?

pylearn2/sandbox/rnn/models/rnn.py
```diff
@@ -189,16 +189,25 @@ def __init__(self, dim, layer_name, irange, indices=None,
...
 
+        # If the unit space is Conv2DSpace, prepare an alternative space
+        # with VectorSpace which dim is equivalent to the original one
+        if isinstance(space.space, VectorSpace):
+            self.requires_reformat = False
+        else:
+            self.requires_reformat = True
+            sequence_space_type = type(self.input_space)
+            self.desired_space = sequence_space_type(
+                VectorSpace(self.input_space.dim))
```
bartvm: Should this not be `self.input_space.get_total_dimension()`?

pylearn2/sandbox/rnn/space/__init__.py
bartvm: I don't see how this would robustly work either, since `data` has an extra dimension for time, that `self.data_space._format_as_impl` would not expect. I think you'd have to merge the time and batch dimensions for this to work.

bartvm:
I didn't have much to add to what @lamblin said. Otherwise, I don't see too much of a problem in merging this; it's sandbox after all.

PR #1460, opened 2015-03-27
Implementing test unit for learning rates

XavierAudier:
Implementing a test for commit #1456


lamblin:
A unit test should check things automatically (and raise an Exception if there is a problem), not simply print things for manual inspection.
If you expect all values to be equal to 1./100, you could for instance call `assert_allclose(..., 1e-2)` on the values, instead of printing them.
Also, you can add the unit test in a different commit of the same pull request, it can make things easier to review.

XavierAudier:
Hi,
Sorry I'll fix that. It's my first time contributing to a repository I'm still not used to it.
Thank you for your patience.

Issue #1172, opened 2014-10-08
speed up NanGuardMode on GPU and move to Theano.


nouiz:
This can be done as in gh-1054. Do the reduction on the GPU, then this will transfer much less data.

The CudaNdarray object do not support many reduction, but we can compile a Theano function that take a gpu object, do the reduction and return the result on the CPU to inspect it.

hantek:
Hi 

I think this ticket is to change something like "np.any(np.isnan(X))" to "np.isnan(np.min(X))", thus to speed up right?

After tracing the method that NanGuardMode uses, I find that it is already using the faster way accordingly, in file /pylearn2/utils/general.py .

lamblin:
Actually, this ticket is about not performing NumPy operations (such as `np.min` and `np.isnan`) when `X` is on GPU (is a `CudaNdarray`, not an ndarray), because in that case, the whole array is first copied on CPU memory as an ndarray, and then `np.min` and `np.isnan` are called.
What we want is that at least the `min` operation runs directly on GPU, and then we can transfer only one number to the CPU.

dwf:
@hantek Any update on this?

hantek:
Hi David

Can we find some codes using the NanGuardMode class? It seems that the constructor doesn't need information about the nodes, but in the __init__() function it needs "node" and "fn" to be the parameter for nan_check().

And, what is a "thunk" in theano?

dwf:
http://deeplearning.net/software/theano/extending/pipeline.html#step-3-execute-linker-to-obtain-a-thunk

You can try this out by compiling a theano function with
`mode=NanGuardMode()`. Just create a function that intentionally gets NaNs
somehow.

hantek:
OK. Thanks a lot!

hantek:
Could you tell me which part of code transfers the data on GPU onto CPU? I looked at the inputs and outputs property of fn, but they are already numpy.ndarray objects. and from other properties of fn I didn't find one with CudaNdarray format. Where can I find the CudaNdarray?

hantek:
I have found one with a more complex computation graph.

hantek:
in the nan_check() function, node.out is a CudaNdarraySharedVariable, which can be used to compute a theano function, and elements in fn.inputs (or fn.outputs) can be both numpy.ndarray or CudaNdarray. 

[...]

PR #1466, opened 2015-04-02
Ignore false positive for GPU_mrg_uniform in nan_guard - fixes #1465

### LINKED ISSUE #1465: NanGuardMode false positive in GPU_mrg_uniform
This resulted from the discussion on [pylearn-users](https://groups.google.com/forum/#!topic/pylearn-users/4E7xYRCJ-7I) mailing list.

Long story short [NanGuardMode](https://github.com/lisa-lab/pylearn2/blob/master/pylearn2/devtools/nan_guard.py) detects Nans with every invocation of GPU_mrg_uniform.

@lamblin provided following insight:
> In more detail, GPU_mrg_uniform actually has an input, which is a shared variable containing the current state of the random number generator. When computing a new sample, the Op actually generates two outputs: a new value for the state, and the sample itself. There is a mechanism to automatically update the value of the shared state to the new expression every time a function using it is called. 
> 
> For MRG, the random state is a matrix of int32. However, when using the GPU, we want the state to be kept and updated in GPU memory for speed, but the current GPU back-end of Theano only supports float32 containers. Therefore, Theano cheats by masquerading a matrix of int32 as a matrix of float32, and the random generator itself knows that it should interpret them as int32. 
>
> The unfortunate side effect is that it looks like a matrix of float32, and that some legitimate values for int32 will have the same binary representation as a NaN in float32, which is probably what is going on in your case.

TNick:
I've also took the liberty to reformat the output a bit and to remove a todo.

### DIFF
--- a/pylearn2/devtools/nan_guard.py
+++ b/pylearn2/devtools/nan_guard.py
@@ -1,14 +1,19 @@
 """
 Functionality for detecting NaNs in a Theano graph.
+
+Some nodes are ignored by name because of known issues:
+- GPU_mrg_uniform: Theano hack (see #1465)
+Please add an explanation if you add nodes to this list.
 """
-__authors__ = "Ian Goodfellow"
+__authors__ = "Ian Goodfellow, Nicu Tofan"
 __copyright__ = "Copyright 2010-2012, Universite de Montreal"
-__credits__ = ["Ian Goodfellow"]
+__credits__ = ["Ian Goodfellow", "Nicu Tofan"]
 __license__ = "3-clause BSD"
 __maintainer__ = "LISA Lab"
 __email__ = "pylearn-dev@googlegroups"
 
 import logging
+import re
 from theano.compile import Mode
 import theano
 import numpy as np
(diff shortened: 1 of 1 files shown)


lamblin:
That's the right place to add the exclusion. However:
- I don't really like the idea of matching types by string.
- It is not all inputs and outputs that should be ignored, just the random states.
I think it may be better to either add a note in the variables' tag, for instance `state.tag.ignore_nan`, when it is built, and then have `NanGuardMode` check that tag.

TNick:
That's a beautiful solution. However, `GPU_mrg_uniform` is not referenced anywhere in pylearn2.
So, you're saying that changes should be made to Theano source, which I'm uncomfortable with doing myself.
Also, based on your comments, the syntax should allow filtering inputs/outputs, not nodes (like in the case of `NanGuardMode` that will  only check generated array and will skip the state.

Bottom line, I believe that some solution should be implemented, as my understanding is that `NanGuardMode` is unusable right now. Am I the only one to use it? Shouldn't be hordes of users complaining?

lamblin:
Yes, I think that the real solution will need a change in Theano, but `NanGuardMode` is in the process of being moved to Theano (see issue #1172), so it does not bother me particularly.
`NanGuardMode` is functional if you do not draw your random numbers on GPU. Since `NanGuardMode` is extremely slow when using the GPU currently (every intermediate result gets transferred to the CPU memory twice), disabling the GPU would be a good idea anyway.

lamblin:
I opened https://github.com/Theano/Theano/issues/2826 as a Theano ticket to take care of it once #1172 is finished.

PR #1472, opened 2015-04-04
Enable monitoring for single model.

hassaku:
When I tried tutorial, the following error occured.

```
$ print_monitor_cv.py cifar_grbm_smd.pkl
Traceback (most recent call last):
  File "XXXXXXX/pylearn2/pylearn2/scripts/print_monitor_cv.py", line 84, in <module>
    main(**vars(args))
  File "XXXXXXX/pylearn2/pylearn2/scripts/print_monitor_cv.py", line 38, in main
    for model in list(this_models):
TypeError: 'GaussianBinaryRBM' object is not iterable
```

The print_monitor_cv.py seems not to support single layer model, so I patched it.
It would be nice if you could adopt this patch.

### DIFF
--- a/pylearn2/scripts/print_monitor_cv.py
+++ b/pylearn2/scripts/print_monitor_cv.py
@@ -35,6 +35,8 @@ def main(models, all=False):
     values = {}
     for filename in np.atleast_1d(models):
         this_models = serial.load(filename)
+        if not isinstance(this_models, list):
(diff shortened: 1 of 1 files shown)


hassaku:
oh... Test is failed. I will check later.

hassaku:
sorry, I should use print_monitor.py not print_monitor_cv.py :bow:

Issue #1088, opened 2014-08-21
Incorporate cost in ConvNonLinearity


lamblin:
The cost should be a default suitable for that non-linearity, and `ConvElemwise.cost` should call the appropriate one, instead of working only on sigmoid.
See https://github.com/lisa-lab/pylearn2/issues/1087

PR #1474, opened 2015-04-08
[WIP] Added tests for comparing mlp and convnet cost functions.

matrogers:


lamblin:
This is for issue #1088

lamblin:
Thanks, that is a good start.
We will see later about making that an actual unit test, and about pep8 and python 3 compliance.
For the moment, I think you are ready to actually start the refactoring (make the cost method part of the NonLinearity class), add an implementation of the default cost for linear (identity), and make the other ones (tanh, rectifier) raise NotImplementedError (like the MLP version).
We will discuss that tomorrow.

PR #1428, opened 2015-03-11
EpochCounter bugfix, fix for #1078

hantek:

PR #1477, opened 2015-04-09
Fix for #1172

hantek:
Test for the presence of 'np.nan', 'np.inf', and bigmnumbers larger than 1e10, using a theano function which finds the minimum/maximun of the theano.sandbox.cuda.CudaNdarray and then transfers the reduced result onto CPU. This implementaion avoids massive data transfer between GPU and CPU when the CudaNdarray is very large, at the expense of compiling an extra theano function at the instantiation of NanGuardMode.

While using CPU, it does exactly the same to its initial version.

### DIFF
--- a/pylearn2/devtools/nan_guard.py
+++ b/pylearn2/devtools/nan_guard.py
@@ -12,6 +12,7 @@ import logging
 from theano.compile import Mode
 import theano
 import theano.tensor as T
+from theano.sandbox.cuda import CudaNdarray
 import numpy as np
 from pylearn2.models.dbm import flatten
 from pylearn2.utils import contains_nan, contains_inf
@@ -37,7 +38,7 @@ class NanGuardMode(Mode):
         If True, raise an error when a value greater than 1e10 is encountered.
     """
     def __init__(self, nan_is_error, inf_is_error, big_is_error=True):
-        self.guard_input = T.vector('nan_guard')
+        self.guard_input = theano.sandbox.cuda.fvector('nan_guard')
         if nan_is_error or inf_is_error:
             self.gpumin = theano.function(
                 [self.guard_input], T.min(self.guard_input),
@@ -75,7 +76,7 @@ class NanGuardMode(Mode):
             error = False
             if nan_is_error:
                 err = False
-                if isinstance(var, theano.sandbox.cuda.CudaNdarray):
+                if isinstance(var, CudaNdarray):
                     err = np.isnan(self.gpumin(var.reshape(var.size)))
                 else:
                     err = contains_nan(var)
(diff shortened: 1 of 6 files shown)


pylearn2/devtools/nan_guard.py
```diff
@@ -36,6 +37,23 @@ class NanGuardMode(Mode):
         If True, raise an error when a value greater than 1e10 is encountered.
     """
     def __init__(self, nan_is_error, inf_is_error, big_is_error=True):
+        self.guard_input = T.vector('nan_guard')
```
nouiz: This force the transfer of the data to the from the GPU to the CPU, then to the GPU again. You should declare this as theano.sandbox.cuda.fvector()

nouiz:
travis tests are failing. I did just a very quick check, not a real review.

hantek:
I cannot reproduce this error on my machine:
>>> AttributeError: 'module' object has no attribute 'CudaNdarray'
Nevertheless I tried to import it explicitly at the beginning of the nan_guard.py file. Hope that could fix this error.

Also Travis is capturing a key error, which stuck me at the previous ticket also #1428 . The weird thing is, nothing is changed for that train_y_roc_auc key, but it turns out to be an error.
>>> KeyError: "'train_y_roc_auc'\n

nouiz:
Travis do not have a gpu. Yet on a computer without a gpu, maybe that will
allow you to reproduce it.

dwf:
This contains commits from unrelated PR, so we'll need to rebase in addition to fixing it such that it skips certain tests when a GPU is unavailable. There's a "skip_if_no_gpu" thing for that.

hantek:
I see, will check that now. Also Mehdi has posted a way to deal with the train_y_roc_auc Key error, I will introduce that into this commit too.

hantek:
Travis is complaining about the reduce() function. Isn't reduce() a built-in function in python? But why we can't use it? 

I will try to do the "git rebase" thing after passing Travis.

hantek:
I find out why. It's because python 3.4 removes that.

dwf:
Reduce is builtin in Python 2. In Python 3 it is a generator, and it lives
in the functools module. You can use six.moves  (Theano bundles a copy of
six, I forget where, check the code) to import it on either.

hantek:
Well... may be I have made this branch more complicated instead... How to remove the unwanted commits by git rebase? (It'd be better if I can talk with you.)

dwf:
Don't forget you'll have to git push --force

hantek:
Yes I did that. There is still an "EpochCounter bugfix" commit, which is unwanted...

dwf:
In that case, what you should do is create a new branch something like this:

```
git fetch origin-lisa
git checkout -b nan2 origin-lisa/master
```

Then use git cherry-pick to get only the good commits:

```
git cherry-pick first_commit_hash_you_want..last_commit_hash_you_want
```

so probably `fd49dd9..c8de035` in this case.

Finally overwrite your old nan branch:

```
git checkout -B nan
```

and `git push --force` again. You can the do `git branch -D nan2` to delete the unneeded second branch.

lamblin:
`git rebase -i` (interactive) also lets you pick only the commits you want.

hantek:
Both of the two ways don't work, git is complaining about one of the patch is conflicting to lisa-labo/pylearn2/master.

hantek:
Hi 
I created a new PR at #1486, now those unwanted commits are removed.

PR #1479, opened 2015-04-13
Wmape channel

junbochen:
add a wmape channel, to calculate wegithed mean average error measure for regression models

### DIFF
--- a/pylearn2/train_extensions/tests/test_wmape_channel.py
+++ b/pylearn2/train_extensions/tests/test_wmape_channel.py
@@ -6,8 +6,7 @@ from pylearn2.testing.skip import skip_if_no_sklearn
 
 
 def test_wmape():
-    """Test RocAucChannel."""
-#     return 
+    """Test WMapeChannel."""
     skip_if_no_sklearn()
     trainer = yaml_parse.load(test_yaml)
     trainer.main_loop()
@@ -16,7 +15,7 @@ def test_wmape():
 test_yaml = """
 !obj:pylearn2.train.Train {
     dataset:
-      &train !obj:pylearn2.testing.datasets.random_dense_design_matrix_for_regression
+        &train !obj:pylearn2.testing.datasets.random_dense_design_matrix_for_regression
       {
           rng: !obj:numpy.random.RandomState { seed: 1 },
           num_examples: 10,
@@ -63,4 +62,3 @@ test_yaml = """
     ],
 }
 """
-
--- a/pylearn2/train_extensions/tests/test_wmape_channel.py
+++ b/pylearn2/train_extensions/tests/test_wmape_channel.py
@@ -1,0 +1,67 @@
+"""
+Tests for ROC AUC.
+"""
+from pylearn2.config import yaml_parse
+from pylearn2.testing.skip import skip_if_no_sklearn
+
+
+def test_wmape():
+    """Test RocAucChannel."""
+#     return 
+    skip_if_no_sklearn()
+    print 'wahaha'
+    trainer = yaml_parse.load(test_yaml)
+    trainer.main_loop()
+
+
+test_yaml = """
+!obj:pylearn2.train.Train {
+    dataset:
+      &train !obj:pylearn2.testing.datasets.random_dense_design_matrix_for_regression
+      {
+          rng: !obj:numpy.random.RandomState { seed: 1 },
+          num_examples: 10,
+          dim: 10,
+          reg_min: 1,
+          reg_max: 1000
+      },
+    model: !obj:pylearn2.models.mlp.MLP {
+        nvis: 10,
+        layers: [
+            !obj:pylearn2.models.mlp.Sigmoid {
+                layer_name: h0,
+                dim: 10,
+                irange: 0.05,
+            },
+            !obj:pylearn2.models.mlp.Linear {
+                layer_name: y,
+                dim: 1,
+                irange: 0.,
+            }
+        ],
+    },
+    algorithm: !obj:pylearn2.training_algorithms.bgd.BGD {
+        monitoring_dataset: {
+            'train': *train,
+        },
+        batches_per_iter: 1,
+        monitoring_batches: 1,
+        termination_criterion: !obj:pylearn2.termination_criteria.And {
+            criteria: [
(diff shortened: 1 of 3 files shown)


junbochen:
why the "CI failed details " is empty: https://travis-ci.org/lisa-lab/pylearn2/builds/58245429

lamblin:
Can you access the individual build pages for the failing parts?
[part1](https://travis-ci.org/lisa-lab/pylearn2/jobs/58245430), [txt](https://s3.amazonaws.com/archive.travis-ci.org/jobs/58245430/log.txt)
[part4](https://travis-ci.org/lisa-lab/pylearn2/jobs/58245433), [txt](https://s3.amazonaws.com/archive.travis-ci.org/jobs/58245433/log.txt)
In any case, the failures are related to pep8 and the documentation standard not being respected in the files you added.

junbochen:
It turns out that the CI build detail page is blocked by the GFW, I could see the details page by using VPN.

pylearn2/train_extensions/wmape_channel.py
```diff
@@ -0,0 +1,116 @@
...
+        node : Apply instance
+            Symbolic inputs and outputs.
+        inputs : list
+            Sequence of inputs.
+        output_storage : list
+            List of mutable 1-element lists.
+        """
+        y_true, y_score = inputs
+        wmape = np.sum(abs(y_true - y_score)) / np.sum(abs(y_true))
+        output_storage[0][0] = theano._asarray(wmape, dtype=config.floatX)
```
lamblin: I'm not sure why you need a new theano Op for that.
Can't you simply use a symbolic expression using existing theano operations, something like `abs(y_true - y_score).sum() / abs(y_true).sum()`, maybe casting `y_true` and `y_score` to `floatX` if needed?

lamblin: This would also make sure the computation is done on the GPU if you are using one, rather than transferring it over to CPU to be able to use numpy.

pylearn2/train_extensions/wmape_channel.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+class WMapeChannel(TrainExtension):
+    """
+    Adds a WMape channel to the monitor for each monitoring dataset.
+
+    This monitor will return nan unless sum of y_true is not 0.
+    For this reason, it is recommended to set monitoring_batches
+    to 1, especially when using unbalanced datasets.
+
```
lamblin: Can you add the mathematical formula in the documentation, as well as a citation if it makes sense?

Actually, this metric does not make that much sense as a monitoring channel, because monitoring channels are computed as the mean of a cost per example, computed minibatch by minibatch.
Even if the sum of y_true is not 0, different batch sizes would give you different results.
You could compute the numerator and the denominator as two different channels, though. I'm not sure what the best way would be to combine them as one number.

lamblin:
Thanks for the update. For the record, we are not notified when new commits are added, only when new comments are published, so adding a comment saying you've updated the PR can help making sure it is addressed.

pylearn2/train_extensions/wmape_channel.py
```diff
@@ -0,0 +1,111 @@
...
+from theano import tensor as T
+
+from pylearn2.train_extensions import TrainExtension
+
+
+class WMapeNumeratorChannel(TrainExtension):
+    """
+    Adds a WMape Numerator channel to the monitor for each monitoring dataset.
+    It calculates the numerator of the WMAPE formula.
+
```
lamblin: Could you put the formula here? It is awkward to have to look for what "the numerator of the WMAPE formula" is.
Also, maybe add a "See also" section mentioning `WMapeDenominatorChannel`.

pylearn2/train_extensions/wmape_channel.py
```diff
@@ -0,0 +1,111 @@
...
+                                      val=wmape_numerator,
+                                      data_specs=(m_space, m_source),
+                                      dataset=dataset)
+
+
+class WMapeDenominatorChannel(TrainExtension):
+    """
+    Adds a WMape Denominator channel to the monitor for each monitoring
+    dataset. It calculates the denominator of the WMAPE formula.
+
```
lamblin: Same here.

pylearn2/train_extensions/tests/test_wmape_channel.py
```diff
@@ -0,0 +1,67 @@
...
+"""
+from pylearn2.config import yaml_parse
+from pylearn2.testing.skip import skip_if_no_sklearn
+
+
+def test_wmape():
+    """Test WMapeChannel."""
+    skip_if_no_sklearn()
+    trainer = yaml_parse.load(test_yaml)
+    trainer.main_loop()
```
lamblin: Can you check here that, after training, the channels have been computed, and have the right value?

lamblin:
When those points are addressed, I think I would be ready to merge.

junbochen:
I've add the WMAPE formula to class comments, and the unit test is also updated

lamblin:
Great, thanks for the update!
Do you think you could try to rebase that branch on top of the current master to remove all the merge commits?
Then you would have to use `git push --force` to update the pull request, in particular not merge with the current branch.
Thanks,

junbochen:
I'm trying to rebase wmape_channel branch to master. However, after the rebase is done, I found the code has been roll back to an old version, in which the unittest code is missing, and the WMapeOp class appears again. 
After doing some google search, I found it's caused by many of the commits in wmape_channel branch is merge commits, and those ones are ignored by git rebase command.
I found the git rebase command provide a '--preserve_merges' option, but while i run the command 'git rebase --preserv_merges --interactive $hash', there's an error message saying 
"error: Commit b7175ad6b9a6a9ff276c34da074d7dc23436204e is a merge but no -m option was given.
fatal: cherry-pick failed
Could not pick b7175ad6b9a6a9ff276c34da074d7dc23436204e"
I searched the problem in google and can't get any help. 
So, could you help me on this issue?

lamblin:
I've had the same problem, I'm not really sure it is possible to make `rebase` work when there are changes in merge commits. There may be a way with `cherry-pick`, but I did not have the patience for that.
In the end, I took the entire diff and put it in only one commit, see #1516.

PR #1480, opened 2015-04-13
Pr 1428

### LINKED ISSUE #1078: termination_criteria.EpochCounter with new_epochs=False does not check before running
I am using `termination_criteria.EpochCounter` with `new_epochs=False` along with `monitor.push_monitor` with `transfer_experience=True`. The model loaded has already seen more than or equal to `max_epochs` but the main training loop still performs one epoch before terminating. 

This is a bug; no training should occur since we are at or exceeding `max_epochs`.

memimo:
It's continuation of @hantek work on #1428 that fixes #1078

### DIFF
--- a/pylearn2/termination_criteria/__init__.py
+++ b/pylearn2/termination_criteria/__init__.py
@@ -12,6 +12,7 @@ __email__ = "pylearn-dev@googlegroups"
 import functools
 import numpy as np
 
+
 class TerminationCriterion(object):
     """
     A callable used to determine if a TrainingAlgorithm should quit
@@ -33,8 +34,8 @@ class TerminationCriterion(object):
             True or False as described above
         """
 
-        raise NotImplementedError(str(type(self)) + " does not implement " +
-                                  "continue_learning.")
+        raise NotImplementedError(
+            str(type(self)) + " does not implement " + "continue_learning.")
 
 
 class MonitorBased(TerminationCriterion):
@@ -55,6 +56,7 @@ class MonitorBased(TerminationCriterion):
         has only one channel, this channel will be used; otherwise, an
         error will be raised.
     """
+
     def __init__(self, prop_decrease=.01, N=5, channel_name=None):
         self._channel_name = channel_name
         self.prop_decrease = prop_decrease
@@ -205,8 +207,7 @@ class EpochCounter(TerminationCriterion):
     Parameters
     ----------
     max_epochs : int
-        Number of epochs (i.e. calls to this object's `__call__`
-        method) after which this termination criterion should
+        Number of epochs after which this termination criterion should
         return `False`.
     new_epochs : bool, optional
         If True, epoch counter starts from 0. Otherwise it
@@ -219,18 +220,17 @@ class EpochCounter(TerminationCriterion):
 
     def initialize(self, model):
         if self._new_epochs:
-            self._epochs_done = 0
+            self._epochs_offset = model.monitor.get_epochs_seen()
         else:
-            self._epochs_done = model.monitor.get_epochs_seen()
+            self._epochs_offset = 0
 
     @functools.wraps(TerminationCriterion.continue_learning)
     def continue_learning(self, model):
-        if not hasattr(self, "_epochs_done"):
+        if not hasattr(self, "_epochs_offset"):
             self.initialize(model)
 
-        old_epochs_done = self._epochs_done
-        self._epochs_done += 1
-        return old_epochs_done < self._max_epochs
+        current_epoch = model.monitor.get_epochs_seen() - self._epochs_offset
+        return current_epoch < self._max_epochs
 
 
 class And(TerminationCriterion):
--- a/pylearn2/termination_criteria/__init__.py
+++ b/pylearn2/termination_criteria/__init__.py
@@ -221,16 +221,16 @@ class EpochCounter(TerminationCriterion):
         if self._new_epochs:
             self._epochs_done = 0
         else:
-            # epochs_seen = 1 on first continue_learning() call
-            self._epochs_done = model.monitor.get_epochs_seen() - 1
+            self._epochs_done = model.monitor.get_epochs_seen()
 
     @functools.wraps(TerminationCriterion.continue_learning)
     def continue_learning(self, model):
         if not hasattr(self, "_epochs_done"):
             self.initialize(model)
 
+        old_epochs_done = self._epochs_done
         self._epochs_done += 1
-        return self._epochs_done < self._max_epochs
+        return old_epochs_done < self._max_epochs
 
 
 class And(TerminationCriterion):
--- a/pylearn2/termination_criteria/tests/test_init.py
+++ b/pylearn2/termination_criteria/tests/test_init.py
@@ -18,10 +18,7 @@ def test_epoch_counter():
     """
     Test epoch counter with max_epochs={True,False}
     """
-
-    N = 5
-
-    def produce_train_obj(new_epochs, model=None):
+    def produce_train_obj(new_epochs, max_epochs, model=None):
         if model is None:
             model = MLP(layers=[Softmax(layer_name='y',
                                         n_classes=2,
@@ -34,7 +31,7 @@ def test_epoch_counter():
         dataset = DenseDesignMatrix(X=np.random.normal(size=(6, 3)),
                                     y=np.random.normal(size=(6, 2)))
 
-        epoch_counter = EpochCounter(max_epochs=N,
+        epoch_counter = EpochCounter(max_epochs=max_epochs,
                                      new_epochs=new_epochs)
 
         algorithm = SGD(batch_size=2, learning_rate=0.1,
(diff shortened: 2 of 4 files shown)


memimo:
@bouthilx or @nouiz could you have a final look on this PR as you guys are more informed with the history of this issues. Thx.

nouiz:
I do not know enough pylearn2 to review this. My first fix I tried was bad. @goodfeli told the problems the old fix tried had. He is in a much better position to review this then me.

goodfeli:
I can review this later today.

pylearn2/train.py
```diff
@@ -140,26 +140,34 @@ def main_loop(self, time_budget=None):
...
-                if self.exceeded_time_budget(t0, time_budget):
-                    break
+            extension_continue = self.run_callbacks_and_monitoring()
+            # First check if the model is already beyond the stop criteria of
+            # training, if so, just return directly.
+            continue_learning = (self.model.continue_learning() and
+                                 extension_continue)
+            assert continue_learning in [True, False, 0, 1]
+            if continue_learning:
+                while True:
```
lamblin: I think it would be cleaner to just have `while continue_learning:` here and drop the `if not continue_learning: break` at the end of the loop.

pylearn2/train.py
```diff
@@ -193,36 +201,47 @@ def main_loop(self, time_budget=None):
...
-                    break
+            # First check if the model is already beyond the stop criteria of
+            # training, if so, just return directly.
+            continue_learning = (
+                self.algorithm.continue_learning(self.model) and
+                extension_continue
+            )
+            assert continue_learning in [True, False, 0, 1]
+            if continue_learning:
+                while True:
```
lamblin: Same here.

lamblin:
I just had a quick look, but did not review in depth.

bouthilx:
I'll take a look this afternoon.

pylearn2/train_extensions/tests/test_roc_auc.py
```diff
@@ -112,7 +112,7 @@ def test_roc_auc_one_vs_one():
                     max_epochs: 1,
                 },
                 !obj:pylearn2.termination_criteria.MonitorBased {
-                    channel_name: train_y_roc_auc,
+                    channel_name: train_roc_auc-0vX,
```
goodfeli: Why does this PR change these channel names?

pylearn2/train.py
goodfeli: +1

pylearn2/train_extensions/tests/test_roc_auc.py
```diff
@@ -173,7 +173,7 @@ def test_roc_auc_one_vs_one():
                     max_epochs: 1,
                 },
                 !obj:pylearn2.termination_criteria.MonitorBased {
-                    channel_name: train_y_roc_auc,
+                    channel_name: train_roc_auc-0v1,
```
hantek: Just curious, if you have time to answer: What happened to train_y_roc_auc? Previously it was there and behaved normally, but after the modifications in train.py it starts to be a Key error. Seems nothing relevant to that is changed in this PR.

pylearn2/termination_criteria/__init__.py
```diff
@@ -221,16 +221,16 @@ def initialize(self, model):
         if self._new_epochs:
             self._epochs_done = 0
```
goodfeli: The docstring for this class needs to be updated, because it defines max_epochs in terms of the number of calls to this class's continue_learning method.

Is the `new_epochs` flag actually useful at all? It seems incompatible with push_monitor. How do you do two-stage learning without push_monitor?

Why not infer the number of epochs seen from monitor.get_epochs_seen() rather than maintaining a separate count of epochs inside this class?

bouthilx: ```new_epochs``` Was to allow counting from new training no matter how many epochs it was trained beforehand. If new_epochs is False, the model is trained while ```max_epochs > old_epochs + epochs_done```. Otherwise, old_epochs is reset to 0 and the model is trained while ```max_epochs > epochs_done```.

pylearn2/termination_criteria/tests/test_init.py
```diff
@@ -48,21 +45,40 @@ def test_epochs(epochs_seen, n):
...
+    train_obj = produce_train_obj(new_epochs=True, max_epochs=5)
+    train_obj.main_loop()
+    test_epochs(train_obj.model.monitor.get_epochs_seen(), 5)
+    train_obj = produce_train_obj(new_epochs=True, max_epochs=5,
+                                  model=train_obj.model)
+    train_obj.main_loop()
+    test_epochs(train_obj.model.monitor.get_epochs_seen(), 2 * 5)
+
+    # Tests for 5 max epochs
+    train_obj = produce_train_obj(new_epochs=False, max_epochs=5)
```
bouthilx: ```new_epochs``` should not make any difference for models trained from scratch. It should be tested that if model is trained again with ```new_epochs=False``` and ```max_epochs=5```, then ```model.monitor.get_epochs_seen() == 5```

pylearn2/termination_criteria/tests/test_init.py
```diff
@@ -48,21 +45,40 @@ def test_epochs(epochs_seen, n):
...
-    test_epochs(train_obj.model.monitor.get_epochs_seen(), N)
+    after_train = train_obj.model.get_weights()
+    test_epochs(train_obj.model.monitor.get_epochs_seen(), 0)
+    assert np.all(before_train == after_train)
 
-    # Try training while already reached max_epochs, should stop after 1 epoch
-    # on first continue_learning() call
-    train_obj = produce_train_obj(new_epochs=False, model=train_obj.model)
+    # Tests for 0 max epochs
+    train_obj = produce_train_obj(new_epochs=False, max_epochs=0)
```
bouthilx: should be 
```python
train_obj = produce_train_obj(new_epochs=False, max_epochs=0,
                              model=train_obj.model)
```

pylearn2/train.py
bouthilx: +1

pylearn2/train_extensions/tests/test_roc_auc.py
memimo: The original test was buggy. The written channel name doesn't exist. The issue was not provoked before, before the EpochCounter terminates the training before even MonitorBased being called.

pylearn2/train_extensions/tests/test_roc_auc.py
memimo: The original test was buggy. The written channel name doesn't exist. The issue was not provoked before, before the EpochCounter terminates the training before even MonitorBased being called.

memimo:
@bouthilx @lamblin could you check a final review, thx!

pylearn2/termination_criteria/__init__.py
lamblin: If `push_monitor` is called with `transfer_experience=True`, then it will work.
Regarding the separate counter, I would agree, counting the number of times `EpochCounter.continue_learning()` has been called is not great, especially if we call it once before the training even starts.

lamblin: Actually, if `new_epochs` is True, then it does not matter how `push_monitor` was called, and it will work as expected.
If `new_epochs` is False, then the result would be different.

pylearn2/train.py
```diff
@@ -140,26 +140,32 @@ def main_loop(self, time_budget=None):
...
-            while True:
-                if self.exceeded_time_budget(t0, time_budget):
-                    break
+            extension_continue = self.run_callbacks_and_monitoring()
+            # First check if the model is already beyond the stop criteria of
+            # training, if so, just return directly.
+            continue_learning = (self.model.continue_learning() and
+                                 extension_continue)
+            assert continue_learning in [True, False, 0, 1]
+            if continue_learning:
```
lamblin: That line should be remove, it is redundant with the "while" on next line.

pylearn2/train.py
```diff
@@ -193,36 +198,45 @@ def main_loop(self, time_budget=None):
...
-                    break
+            # First check if the model is already beyond the stop criteria of
+            # training, if so, just return directly.
+            continue_learning = (
+                self.algorithm.continue_learning(self.model) and
+                extension_continue
+            )
+            assert continue_learning in [True, False, 0, 1]
+            if continue_learning:
+                while continue_learning:
```
lamblin: Idem here, the "if" and "while" are redundant

lamblin:
Thanks for addressing the comments.
Does any other reviewer have outstanding comments?

pylearn2/termination_criteria/__init__.py
bouthilx: @lamblin Do you think it's correct? If they use transfer_experience=False they should not want to preserve previous number of epochs so new_epochs=False wouldn't make sense and will behave as new_epochs=True anyway.

bouthilx:
Looks good. I left a comment regarding the coupled behavior of ```push_monitor(transfer_learning)``` and ```MonitorBased(new_epochs)```. @lamblin If you think it's OK then I agree for merging.

pylearn2/termination_criteria/__init__.py
lamblin: Right, the behaviour is:
- `new_epochs=True`: start at 0, regardless of `transfer_experience`
- `transfer_experience=False`: start at 0, regardless of `new_epochs`
- `new_epochs=False` and `transfer_experience=True`: start at the previous number of epochs.
It makes sense to me at least, do you think it is confusing? Would you expect something different?

bouthilx: Makes sense to me.

Issue #1126, opened 2014-09-04
Fix max col norm formula


goodfeli:
Currently max col norm / kernel norm stuff uses (variations on) this formula:

updates[W] = updated_W * (desired_norms / (1e-7 + col_norms))

Adding 1e-7 causes the parameters to shrink gradually. This is usually not noticeable but it is very noticeable if the weights are small and you set the learning rate to 0 for debugging purposes. This could be fixed by using T.switch to set the scale factor to 1 for columns that are within the allowed norm range.

I think @lamblin is going to make a separate ticket to unify the max col norm implementations with a single ModelExtension plugin, so it could make sense to do that first so there is only one piece of code to fix.

lamblin:
The refactoring of max col norm is not happening soon, so let's not wait to address this one.

robintibor:
One question regarding this.
I saw in the pylearn terminal output during trianing that the actual max col norm for my softmax layer was exceeding the max col norm constraint that i specified (actual max col norm typically between 0.6 and 1.0, once even at 1.7, specified max_col_norm was 0.5) .
Is it intended that it's
```python 
col_norms = T.sqrt(T.square(W).sum(axis=0)) 
```
in https://github.com/lisa-lab/pylearn2/blob/master/pylearn2/model_extensions/norm_constraint.py#L47

and not 
```python 
col_norms = T.sqrt(T.square(updated_W).sum(axis=0)) 
```
as in https://github.com/lisa-lab/pylearn2/blob/master/pylearn2/models/mlp.py#L2045 ?

If it is correct this way, can I just ignore the output on the terminal saying the max col norm is too high? Hope I am not missing something here :)

lamblin:
Using the norm of `updated_W` would make more sense to me.
@goodfeli can you confirm?

goodfeli:
Yes, it should be updated_W, sorry.

2015-04-23 12:00 GMT-07:00 Pascal Lamblin <notifications@github.com>:

lamblin:
Thanks.
@laurent-dinh can you take care of that as well?

goodfeli:
I already sent a PR for the updated_W bug:
https://github.com/lisa-lab/pylearn2/pull/1498

2015-04-23 12:14 GMT-07:00 Pascal Lamblin <notifications@github.com>:

lamblin:
Fixed in #1484.

PR #1484, opened 2015-04-15
Fixing the norm constraint formula

laurent-dinh:


JesseLivezey:
Would it make sense to factor out some of the computation into a ```pylearn2.expr.updates.py``` file or something similar so that the same/similar code isn't rewritten everywhere? This might also make testing easier.

lamblin:
I agree with @JesseLivezey, could you factor it out? Especially if we want to later change the 1e-7 magical constant.

memimo:
Looks like his new commit is not triggering travis. @lamblin do you know if what could be the cause?

lamblin:
It might have been an issue of Travis itself. I'll close and re-open.

lamblin:
Putting just the `max` is not that helpful... I would have expected something like:
```python
def renormalize_factor(desired_norms, norms, eps=1e-7):
    """
    (Helpful docstring here)
    """
    return desired_norms / T.maximum(norms, eps)
```
That way, we have more control over `eps` as well.

goodfeli:
I already fixed this, in a factored way, for maxout. Please use the same MaxL2FilterNorm ModelExtension here too.

goodfeli:
Here's the ModelExtension: https://github.com/lisa-lab/pylearn2/blob/c4293f99fa2028cc486512b8989e9c5c5fb87132/pylearn2/model_extensions/norm_constraint.py

goodfeli:
Here's how I use it for maxout: https://github.com/lisa-lab/pylearn2/blob/master/pylearn2/models/maxout.py#L172

laurent-dinh:
Thank you Ian.
Although I am well aware of this ModelExtension, I am currently trying to address [this issue](https://github.com/lisa-lab/pylearn2/issues/1126) in this PR. As this PR is turning into refactoring the norm constraint, it's good that you brought that into the topic. Should we close this PR and focus on modifying and using this ModelExtension for all concerned models ?

goodfeli:
Well, 1126 does say to do 1128 first.
Basically if you do 1128, then 1126 is no longer relevant because the ModelExtension is already fixed.

lamblin:
@laurent-dinh OK, it would be better to use the existing ModelExtension, I forgot that it already existed.

pylearn2/models/mlp.py
```diff
@@ -2024,15 +2020,6 @@ def _modify_updates(self, updates):
...
-        if self.max_row_norm is not None:
-            W, = self.transformer.get_params()
-            if W in updates:
-                updated_W = updates[W]
-                row_norms = T.sqrt(T.sum(T.sqr(updated_W), axis=1))
-                desired_norms = T.clip(row_norms, 0, self.max_row_norm)
-                scales = desired_norms / (1e-7 + row_norms)
-                updates[W] = updated_W * scales.dimshuffle(0, 'x')
-
         if self.max_col_norm is not None or self.min_col_norm is not None:
```
laurent-dinh: Should I integrate a minimum norm constraint ?
If so, should I forbid using row and column norm constraint at the same time.

memimo: yeah, let's integrate the minimum norm constraint.
And yes, you should forbid both row and column norm constraints at same time, that's what current code is doing now.

goodfeli:
The minimum norm issue is separate from the row norm issue.

As far as I know, no one uses either max_row_norm or min_col_norm. Ideally, we'd mark them as deprecated for 6 months before removing them. In this case it might be OK to just remove them.

I wrote them while experimenting with what kind of norm constraints work well, and after deciding that max_col_norm was the only one that was consistently helpful I have quit using the other two. I don't know of anyone else using them.

pylearn2/model_extensions/norm_constraint.py
```diff
@@ -21,12 +21,19 @@ class MaxL2FilterNorm(ModelExtension):
     Parameters
     ----------
     limit : float or symbolic float
-        The maximum column norm of the weight matrix is constrained
-        to be <= limit
+        The maximum norm of the weight matrix is constrained
+        to be <= limit along the axes
+    limit : float or symbolic float
```
memimo: min_limit*

pylearn2/model_extensions/norm_constraint.py
```diff
@@ -21,12 +21,19 @@ class MaxL2FilterNorm(ModelExtension):
...
+        The maximum norm of the weight matrix is constrained
+        to be => limit along the axes
```
memimo: minimum

memimo:
Except the minor doc comments LGTM.

laurent-dinh:
Alright, the Travis error caused by an external library should be fixed now. 
Can someone relaunch the Travis test ?

laurent-dinh:
Thanks !

memimo:
@lamblin if you have no further comment, I will merge today.

pylearn2/model_extensions/norm_constraint.py
```diff
@@ -21,12 +21,19 @@ class MaxL2FilterNorm(ModelExtension):
...
+    axis : int or tuple of int
+        The axis or axes over which the norm is computed. Default is 0.
     """
 
-    def __init__(self, limit):
-        self.limit = limit
+    def __init__(self, limit, min_limit=0., axis=0):
+        self.max_limit = limit
+        self.min_limit = min_limit
+        self.axis = axis
```
lamblin: If both `min_limit` and `max_limit` are provided, can you assert that `min_limit <= max_limit`? The behaviour in the other case is ill-defined.

lamblin:
Other than that, it works for me.

laurent-dinh:
The modifications asked are done and the build passed.

PR #1491, opened 2015-04-21
Updated test based on Pascal Lamblin's comments

jovallee:
Changed the integer division to integer / float
Changed the L1WeightDecay(tuple) to L1WeightDecay(dict)

### DIFF
--- a/pylearn2/costs/tests/test_term_1_0_l1_penalty.py
+++ b/pylearn2/costs/tests/test_term_1_0_l1_penalty.py
@@ -59,8 +59,7 @@ def test_correctness():
                                  N=2)])
 
     cost = SumOfCosts([(0.99, Default()),
-                       (0.01, L1WeightDecay({'hidden': 0.,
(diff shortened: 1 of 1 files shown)


pylearn2/costs/tests/test_term_1_0_l1_penalty.py
```diff
@@ -59,7 +59,8 @@ def test_correctness():
                                  N=2)])
 
     cost = SumOfCosts([(0.99, Default()),
-                       (0.01, L1WeightDecay((0., 1.0)))])
+                       (0.01, L1WeightDecay({'hidden': 0.,
+                                             'output': 1.0}))])
```
lamblin: Oh, but if you actually pass the parameters in a dictionary, it will not test the first bug anymore. You would have to pass an empty dictionary for that.

jovallee:
Thanks for taking the time to help me Pascal.

lamblin:
Tanks for the update.

PR #1497, opened 2015-04-23
Raising a `RuntimeError` when NaN is found during training

alexjc:
When using PyLearn2 from another python library, I've found it would be easier to have specific exceptions to deal with common problems during training (e.g. NaN in SGD).

This PR will raise a `RuntimeError` instead of an exception when training results in NaN. This makes it easier to catch problems and fits a little better with the official Python exception classifications.


lamblin:
That seems reasonable, thanks!

Issue #1462, opened 2015-03-31
[enhancement] Consistent parameter naming


TNick:
A network with three `maxout.MaxoutLocalC01B`, a `maxout.Maxout` and a `mlp.Softmax` prints `Parameter and initial learning rate summary` like so:
```
	W: 0.0025
	b: 0.0025
	W: 0.0025
	b: 0.0025
	W: 0.0025
	b: 0.0025
	layer_4_W: 0.05
	layer_4_b: 0.05
	softmax_b: 0.05
	softmax_W: 0.05
```
This is all but consistent. I like the `layer.name + W/b` notation.
Would you accept a PR with that? If not, please suggest other naming convention and I would be happy to write a PR.

lamblin:
Yes, that looks like a good idea, a PR would be welcome.
Thanks!

TNick:
`softmax_b` and `softmax_W` too?
```
$>grep -R softmax_b
models/dbm/layer.py:        self.b = sharedX( np.zeros((n_classes,)), name = 'softmax_b')
models/mlp.py:                             name='softmax_b')
scripts/tutorials/convolutional_network/convolutional_network.ipynb:        "\tsoftmax_b: 0.00999999977648\n"
scripts/tutorials/jobman_integration.ipynb:        "\tsoftmax_b: 0.000205\r\n",
scripts/tutorials/multilayer_perceptron/multilayer_perceptron.ipynb:        "\tsoftmax_b: 0.00999999977648\n"
scripts/tutorials/multilayer_perceptron/multilayer_perceptron.ipynb:        "\tsoftmax_b: 0.00999999977648\n"
scripts/tutorials/stacked_autoencoders/stacked_autoencoders.ipynb:        "\tsoftmax_b: 0.0500000007451\n"
```
```
$>grep -R softmax_W
models/dbm/layer.py:        self.W = sharedX(W,  'softmax_W' )
models/mlp.py:            self.W = sharedX(W,  'softmax_W')
scripts/tutorials/convolutional_network/convolutional_network.ipynb:        "\tsoftmax_W: 0.00999999977648\n"
scripts/tutorials/jobman_integration.ipynb:        "\tsoftmax_W: 0.000205\r\n"
scripts/tutorials/multilayer_perceptron/multilayer_perceptron.ipynb:        "\tsoftmax_W: 0.00999999977648\n"
[...]

PR #1500, opened 2015-04-26
Consistent layer naming - addresses #1462

TNick:
The commit also changes some IPython notebooks.
Layers that were named "y" are now named "softmax" to generate same output as before.


lamblin:
I'm not sure about the changes to the .ipynb files, since renaming the layer from `'y'` to `'softmax'` will change all channels that were named `'test_y_...'`, for instance `'test_y_misclass'`. There seem to be more of them than `'softmax_...'` channels, so I would do the changes the other way around.

TNick:
Right. So - by assigning the issue to yourself - you mean that you're going to fix it or you want me to create a new pull request?

lamblin:
Sorry for the misunderstanding, since GitHub only lets us assign people with commit rights to the repository, we have started assigning to PRs the person who will be responsible for reviewing/merging them. It also makes it easier for me to have an overview of the PRs and tickets I should review.
If you could update this PR (I don't think you need to create a new one, if there is a conflict you can always `git rebase` and `git push -f` to the same branch), that would be great. Thanks!

PR #1502, opened 2015-04-26
tear_down for train extensions; fixes #1467

### LINKED ISSUE #1467: [enhancement] tear_down() method for train extensions
[TrainExtension](https://github.com/lisa-lab/pylearn2/blob/master/pylearn2/train_extensions/__init__.py#L16) has a `setup()` but not a `tear_down()` /  `end()` / `terminate()` / `whatever()`. That's probably because it is not needed most of the time.

One use case for it would be with the [LiveMonitoring](https://github.com/lisa-lab/pylearn2/blob/master/pylearn2/train_extensions/live_monitoring.py#L141) that binds to ports but never unbinds. It can be easily re-factored to bind in `setup()` and unbind in `tear_down()`.
(one may argue that, in an yaml file, LiveMonitoring instance can be shared).

The implementation would change [train.py](https://github.com/lisa-lab/pylearn2/blob/master/pylearn2/train.py):
```python
    def tear_down(self):
        """
        Called at the end of main loop.
        """
        self.tear_down_extensions()
        # Does the algorithm also needs one?
        #if self.algorithm is not None:
        #    self.algorithm.tear_down()
            
    def tear_down_extensions(self):
        """ Calls tear_down on all extensions."""
        for ext in self.extensions:
            ext.tear_down(self.model, self.dataset, self.algorithm)
```
`self.tear_down()` would then be called at the end of main loop.

------

Bonus points
------------------
How would you feel about wrapping the main loop in a try ... except like so:
```python
    def main_loop(self, time_budget=None):
        t0 = datetime.now()
        self.setup()
        try:
            if self.algorithm is None:

            # ...

            if self.save_freq > 0:
                self.save()
        except:
            raise
        finally:
            self.tear_down()
```
?

TNick:
`Train` main loop is wrapped in a try-catch block.
`LiveMonitoring` now uses `setup()` to open the ports and `tear_down()` to close them.

### DIFF
--- a/pylearn2/train.py
+++ b/pylearn2/train.py
@@ -94,6 +94,15 @@ class Train(object):
         for ext in self.extensions:
             ext.setup(self.model, self.dataset, self.algorithm)
 
+    def tear_down_extensions(self):
+        """ Calls tear_down on all extensions."""
+        for ext in self.extensions:
+            try:
+                ext.tear_down(self.model, self.dataset, self.algorithm)
+            except Exception:
+                log.debug('%s train extension failed to terminate gracefully',
+                          exc_info=True)
+
     def exceeded_time_budget(self, t0, time_budget):
         """
         .. todo::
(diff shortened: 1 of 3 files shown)


lamblin:
Thanks for the PR.
Unfortunately, I was not able to review it in time, and you will have to rebase it, sorry about that.

Issue #1496, opened 2015-04-23
AdaGrad rule yields NaN on first iteration


alexjc:
I'm running simple tests on small networks for variety of learning rules, and am finding that AdaGrad currently doesn't seem to work on a couple machines I tried:

```
Traceback (most recent call last):
  File "pylearn2/training_algorithms/sgd.py", line 470, in train
    raise Exception("NaN in " + param.name)
Exception: NaN in Output_Linear_W
```

This is a linear layer, but the problem seems to happen regardless of layer type.

Am I missing anything?

TNick:
Try using [NanGuard][1] to detect where it gets produced? Be aware of #1465.

  [1]: https://github.com/lisa-lab/pylearn2/blob/master/pylearn2/devtools/nan_guard.py

alexjc:
OK, here's what I got:

```
Traceback (most recent call last):
  ...
  File "pylearn2/pylearn2/training_algorithms/sgd.py", line 455, in train
    self.sgd_update(*batch)
  File "/Library/Python/2.7/site-packages/theano/compile/function_module.py", line 595, in __call__
    outputs = self.fn()
  File "/Library/Python/2.7/site-packages/theano/gof/link.py", line 797, in f
    raise_with_op(node, *thunks)
  File "/Library/Python/2.7/site-packages/theano/gof/link.py", line 795, in f
    wrapper(i, node, *thunks)
  File "/Library/Python/2.7/site-packages/theano/gof/link.py", line 810, in wrapper
    f(*args)
  File "pylearn2/pylearn2/devtools/nan_guard.py", line 105, in nan_check
    do_check_on(x, node, fn, False)
  File "pylearn2/pylearn2/devtools/nan_guard.py", line 84, in do_check_on
    assert False
AssertionError:
Apply node that caused the error: Elemwise{true_div,no_inplace}(DimShuffle{x,x}.0, Elemwise{sqrt,no_inplace}.0)
Inputs types: [TensorType(float32, (True, True)), TensorType(float32, matrix)]
Inputs shapes: [(1, 1), (16, 4)]
Inputs strides: [(4, 4), (16, 4)]
Inputs values: [array([[ -9.99999997e-07]], dtype=float32), 'not shown']

Backtrace when the node is created:
  File "pylearn2/pylearn2/training_algorithms/learning_rule.py", line 367, in get_updates
    * grads[param])

[...]

PR #1504, opened 2015-04-27
AdaGrad accepts zero-filled params - fixes #1499, #1496

### LINKED ISSUE #1499: Use minimal value for the denominator in AdaGrad
See #1496 for a corner case where the accumulated gradient can be 0.

TNick:
Also enhances `DummyModel` to initialize the parameters to either zeros or random values.
`test_adagrad()` is modified to allow for a factored-out initialization of the test that is then reused in `test_adagard_max_scaling()`.

### DIFF
--- a/pylearn2/training_algorithms/tests/test_learning_rule.py
+++ b/pylearn2/training_algorithms/tests/test_learning_rule.py
@@ -3,6 +3,7 @@ import numpy as np
 from theano.compat.six.moves import zip as izip
 
 from pylearn2.costs.cost import SumOfCosts
+from pylearn2.datasets.dense_design_matrix import DenseDesignMatrix
 from pylearn2.testing.cost import SumOfOneHalfParamsSquared
 from pylearn2.testing.cost import SumOfParams
 from pylearn2.testing.datasets import ArangeDataset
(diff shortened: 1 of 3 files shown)


alexjc:
I've tested this manually against #1496 too and it works for that more complex case as well!

I looked at the code as a relative PyLearn2 beginner, and brought up two comments — but make of those what you will.  The rest looks OK to me.

Great work @TNick!

TNick:
Thanks for the input and testing. See my replies to your comments.

alexjc:
No problem. Thanks for fixing it so quickly!

pylearn2/training_algorithms/tests/test_learning_rule.py
```diff
@@ -170,23 +170,34 @@ def adadelta_manual(model, state):
                izip(manual, model.get_params()))
 
 
-def test_adagrad():
+def prepare_adagrad_test(dataset_type='arrange', model_type='random'):
```
lamblin: Maybe "arange" instead of "arrange", it is named after `numpy.arange` (for "array" and "range"), not after the English word [arrange](http://en.wiktionary.org/wiki/arrange).

lamblin:
Apart from that minor thing, it looks good to me.
Thanks!

PR #1511, opened 2015-05-03
Added a routine to ForcedEvenIterator that is needed to make it work wit...

daemonmaker:
...h the data transformer.


daemonmaker:
Hi @lamblin, I ran into trouble using ForcedEvenIterator with TransformerData when using iteration mode "even_sequential". Would you mind looking over these changes and merging this request if you agree that it's the correct approach?

lamblin:
Yes, it makes sense.
Thanks!

PR #1512, opened 2015-05-04
Rebase of #1486

### LINKED ISSUE #1172: speed up NanGuardMode on GPU and move to Theano.

dwf:
Fixes #1172.

Already reviewed, merge at will when Travis passes.

### DIFF
--- a/pylearn2/train_extensions/tests/test_roc_auc.py
+++ b/pylearn2/train_extensions/tests/test_roc_auc.py
@@ -112,7 +112,7 @@ test_yaml_ovr = """
                     max_epochs: 1,
                 },
                 !obj:pylearn2.termination_criteria.MonitorBased {
-                    channel_name: train_roc_auc-0vX,
+                    channel_name: train_roc_auc_0vX,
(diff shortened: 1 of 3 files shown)


pylearn2/train_extensions/tests/test_roc_auc.py
```diff
@@ -112,7 +112,7 @@ def test_roc_auc_one_vs_one():
                     max_epochs: 1,
                 },
                 !obj:pylearn2.termination_criteria.MonitorBased {
-                    channel_name: train_roc_auc-0vX,
+                    channel_name: train_roc_auc_0vX,
```
lamblin: This seems to trigger the following error in Travis: `KeyError: "'train_roc_auc_0vX'`

dwf:
Fixed the stupid channel name issue and Travis passes.

PR #1514, opened 2015-05-06
Cleaned up documentation by removing unnecessary "WRITEME"s.

daemonmaker:


daemonmaker:
@lamblin I removed the WRITEMEs because there is a "Notes" section that states that the associated properties have the same meaning as in the autoencoder class.

lamblin:
Thanks! I think it's fine to refer to the documentation of the other class, rather than have to copy and paste it, or explicitly manipulate the docstring.

PR #1519, opened 2015-05-08
Upgraded weight initialization to include initializing by sampling a …

daemonmaker:
…Gaussian.

@lamblin I tried to make this backwards compatible however it's not 100% so because I added a parameter in the middle of some of the parameter lists. I can move that to the end if that's preferable.

### DIFF
--- a/pylearn2/models/autoencoder.py
+++ b/pylearn2/models/autoencoder.py
@@ -232,15 +232,14 @@ class Autoencoder(AbstractAutoencoder):
             assert istdev is None
             W = rng.uniform(
                 -irange,
-                 irange,
-                 (nvis, self.nhid)
-                 )
+                irange,
+                (nvis, self.nhid)
+            )
         else:
             assert istdev is not None
             W = rng.randn(nvis, self.nhid) * istdev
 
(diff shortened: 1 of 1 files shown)


daemonmaker:
@lamblin Travis is failing but the error doesn't appear to be related to my changes. Would you please look at the error and let me know whether it is something I need to chase done to get this merged?

daemonmaker:
@lamblin It appears Travis is passing now though I didn't change anything. Can this be merged?

bartvm:
Te test that failed has some randomness in it, and you just got unlucky and fell right above the tolerance level, so restarting it solved the problem.

lamblin:
@bartvm Do you remember which test that was?
@daemonmaker It looks fine, thanks!

bartvm:
I'm quite sure it was `test_zca_dataset`, one of [these two lines](https://github.com/lisa-lab/pylearn2/blob/master/pylearn2/datasets/tests/test_zca_dataset.py#L35-L36) probably, because I can remember `atol` being set to something non-zero.

lamblin:
@bartvm Thanks, I'll try to make the test deterministic again.

PR #1522, opened 2015-05-13
Fix yaml files in scripts/papers/maxout

lamblin:
- changed models to not require fixed batch, remove use of forced even iterators
- fix padding size in cifar scripts with data augmentation

This should address the problems reported in https://groups.google.com/d/topic/pylearn-dev/geKKzmQ1QbM/discussion


goodfeli:
Looks good to me, thanks! Is it OK to merge it?

lamblin:
Sure, go ahead!

PR #1535, opened 2015-06-03
Load correct targets when split is 'train_all'

### LINKED ISSUE #1532: SVHN bugs
~~`which_set='splitted_train'` and `which_set='valid'` fail because `split_train_valid` returns [four values, not two](https://github.com/lisa-lab/pylearn2/blob/393a75583e20b7ec7a539bdde66eb06596afe90d/pylearn2/datasets/svhn.py#L238-239). There's an easy fix, which is to simply turn the first two as a tuple and the last two as a tuple, because `split_train_valid` used to do this and apparently [the relevant code still expects this](https://github.com/lisa-lab/pylearn2/blob/393a75583e20b7ec7a539bdde66eb06596afe90d/pylearn2/datasets/svhn.py#L248).~~

Also, [this](https://github.com/lisa-lab/pylearn2/blob/393a75583e20b7ec7a539bdde66eb06596afe90d/pylearn2/datasets/svhn.py#L260-261) looks totally suspicious and I suspect that `which_set='train_all'` has incorrect labels.

lamblin:
Move assertions at a point they would be able to detect that.

Fixes #1532 (what is remaining of it).

### DIFF
--- a/pylearn2/datasets/svhn.py
+++ b/pylearn2/datasets/svhn.py
@@ -265,16 +265,16 @@ class SVHN(dense_design_matrix.DenseDesignMatrixPyTables):
             if which_set == 'train_all':
                 train_x, train_y = load_data("{0}train_32x32.mat".format(path))
                 data_x = numpy.concatenate((data_x, train_x))
-                data_y = numpy.concatenate((data_y, data_y))
+                data_y = numpy.concatenate((data_y, train_y))
+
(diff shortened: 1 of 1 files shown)


fishcorn:
I probably closed the issue too early, but the fix is pretty obvious so it didn't seem like a big deal.

memimo:
Thanks for this fix.

PR #1539, opened 2015-06-16
Stop using the bundled six that comes with Theano.

abergeron:
We want to remove the bundled copy in Theano so this makes pylearn2
compatible with that future change.


lamblin:
.requirements.txt should also be updated, I think. No need to update the "requirements" part of the doc since "dependencies of Theano" are already mentioned.

nouiz:
I'm not sure this is good as told in the python 2 and 3 PR in Theano.

PR #1540, opened 2015-06-17
Fix typo preventing symbols from being exported

lamblin:


lamblin:
I'm running the GPU tests locally, I'll update if they pass.

nouiz:
You can merge when your local tests pass.

lamblin:
I can confirm that my local tests reported the problem without the fix, and pass with it.

PR #1543, opened 2015-06-24
remove NanGuardMode

hantek:
NanGuardMode removed in this PR since it is now part of theano.


hantek:
This is part of a fix for https://github.com/Theano/Theano/issues/3034

pylearn2/devtools/nan_guard.py
```diff
@@ -1,143 +0,0 @@
-"""
```
nouiz: Keep that file. Make it print a deprecation warning at import. Make it import the Theano version of nanguardmode so that old code continue to work for some time.

dwf: @hantek Once you do this (should be 3 lines) it'll be ready for merge.

hantek:
Hi, so I have brought the file back. Now it's importing from theano's NanGuardMode and printing a depreciation warning.

nouiz:
thanks

PR #1546, opened 2015-06-28
ICML 2013 emotions dataset fix: use next() function for CSV reader, p…

avtomaton:
…roper dataset values extraction, map(...) -> list comprehension (python3 issue)


avtomaton:
There are some errors in python3 for icml_2013_wrepl/emotions - hope my changes can be helpful. Without these fixes I can't even create working softmax model due to exceptions.

PR #1513, opened 2015-05-06
Issue1088

matrogers:
Nearly complete.

PR #1551, opened 2015-07-10
Conv nonlinearity cost

### LINKED ISSUE #1088: Incorporate cost in ConvNonLinearity

lamblin:
Rebase of #1513 and #1547 by @matrogers.
Fixes #1088.


pylearn2/models/mlp.py
```diff
@@ -2734,6 +2764,11 @@ def apply(self, linear_response):
             linear_response * (linear_response < 0.)
         return p
 
+    @wraps(ConvNonlinearity.cost)
+    def cost(self, Y, Y_hat, batch_axis):
+        raise NotImplementedError(
+            str(type(self)) + " does not implement cost function.")
```
bouthilx: Isn't the same as ConvNonlinearity.cost? Why overwriting it then?

lamblin: Right, we can remove that.

pylearn2/models/tests/test_convelemwise_cost.py
```diff
@@ -0,0 +1,172 @@
...
+                                num_channels=1),
+        layers=[ConvElemwise(layer_name='conv',
+                             nonlinearity=conv_nonlinearity,
+                             output_channels=1,
+                             kernel_shape=[1, 1],
+                             pool_shape=[1, 1],
+                             pool_stride=[1, 1],
+                             irange=1.0)],
+        batch_size=1
+    )
```
lamblin: This currently fails, we should take code from the other `check_` method instead.
Also, we should check that the output is the same, even if the cost is not implemented.

lamblin:
The new tests pass, and I fixed an issue with conda stalling when trying to solve the dependencies for package installation.
The remaining issues are due to the fact that pylearn2 depends on an un-maintained copy of `scan` that was in `theano.sandbox`, and that has been removed from the development version.

nouiz:
This PR changed groundhog to use the current scan by just changing the import:

https://github.com/lisa-groundhog/GroundHog/pull/35

This can probably be done here too.

lamblin:
Travis is passing now.

PR #1568, opened 2015-12-17
[RFC] Removing entries with DBM in test_format.py

souravsingh:


nouiz:
travis tests are failing. This can't be merged before merging:

https://travis-ci.org/lisa-lab/pylearn2/jobs/974979189

Do you know about the warning on pylearn2? https://github.com/lisa-lab/pylearn2/

There is no more main developer of pylearn2.

souravsingh:
I am checking it out.I think the errors are something to do with the PR

nouiz:
Yes. By default there is a test that check file formating of all files in
Pylearn2. this test have a white list. You removed files from it, but you
didn't fixed the formating issue in those files.

PR #1573, opened 2016-04-03
Try to fix travix install as was done in Theano

nouiz:


Prateek2901:
Or we can use a simple command to install  **Theano** 

`pip install Theano`

PR #1578, opened 2017-01-31
Make cuda-convnet wrapper compatible to recent Theano

f0k:
The cuda-convnet wrapper uses `make_thunk` to check if cuda-convnet is available. https://github.com/Theano/Theano/pull/5073 changed the `make_thunk` signature to include an `impl` parameter. This PR changes all `make_thunk` calls to pass on `*args` and `**kwargs` so it is compatible to all versions of Theano. (Theano also has a `prepare_node` method which would be suitable for this check, but I'm not sure if this would work with older Theano versions then.)


nouiz:
I'm merging as this PR is good even if travis fail. Travis fail as it install python 2.6 and Theano don't support it anymore.

PR #1581, opened 2017-02-14
fixed the "ImportError: cannot import name string"

Coderx7:
Changed and fixed the wrong import `from pylearn2.utils import string` to its correct form of `from pylearn2.utils import string_utils as string`


nouiz:
travis failure are not related, so merging. Theano don't support python 2.6 anymore.

PR #1583, opened 2017-03-13
Fixes dimension reduction and add Unit Test

t0nyt93:


pylearn2/datasets/preprocessing.py
```diff
@@ -666,7 +666,7 @@ def apply(self, dataset, can_fit=True):
         """
         X = dataset.get_design_matrix()
         if can_fit:
-            self._mean = X.mean(axis=self._axis)
+            self._mean = X.mean(axis=self._axis, keepdims=True)
```
nouiz: This change the behavior of this class. This can break codes. If you want to do that, add a keepdims param in the init that default to False to don't change the behavior.

Note, Pylearn2 isn't supported anymore, so I won't merge PR that change the behavior event if this is a good long term idea, as there is no long term plan. But I can merge PR that help user execute the old code to reproduce old experiments. Nothing more.

PR #1590, opened 2017-11-04
self.tied_b has been set to True

MuriloMir:


nouiz:
Pylearn2 isn't maintained anymore. I think nobody will review this PR. I don't feel save to merge this change of default behavior.

PR #1596, opened 2020-02-03
Merge pull request #3 from lisa-lab/master

Coderx7:
fixed the ImportError: cannot import name string
(This is a 3 years old patch, which I just noticed hasnt been PRed!)


nouiz:
There is no changes in this PR.
Also Pylearn2 isn't maintained for years.... The CI is brokens. So I would be very reluctant to merge anything.

Coderx7:
@nouiz : ah, my bad! I made gaf it seems! (also it seems the prs were already merged back in the day)
I'm sorry for the nuisance here. 
Have a great day

